PR #7, opened 2015-07-31
optimization pull request.

wangxiao1254:
speed now is slightly higher than 5M AND gates/sec. 

it runs correctly with edit distance and benchmark programs


samee:
Actual speed still depends on application, but yes, these seem to make things about 10-15% faster across the board, on fast networks.

PR #8, opened 2015-08-06
Tutorial, linear regression implementation

havron:
obliv-c/tutorial/olinReg has not been able to compile when using the newest CIL version; it has only successfully compiled and ran using the older CIL version that my fork is using.


samee:
Unfortunately, I cannot merge code into the new version if it does not compile with the new version. The errors right now are caused by two issues:

  * Obliv-C no longer supports the use of `yaoGateCount()` after the protocol has finished. So just disabling that one line fixes the "Undefined function" error.
  * CIL seems to be generating "goto" commands whenever you use logical `&&`, which is causing problems. This is a bug on my side, which I will fix. But for now, you can just change the Makefile to compile linReg.c separately with `gcc -c linReg.c` and just link it in. 

This will fix the issue for now.

PR #11, opened 2016-06-27
Close memory leak in getsockaddr

schoppmp:
Fixes a memory leak in `getsockaddr` that arises when the first element in the list allocated by `freeaddrinfo` does not satisfy `.ai_family == AF_INET`.


samee:
Thanks Phillipp! Merged.

PR #12, opened 2016-09-24
Added compatibility with Fedora

jackdoerner:
Changed curve used for OT from secp192r1 to secp256r1.
Updated README with dependencies for Fedora (and other RHEL derivates).

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,13 @@ Obliv-C is a simple GCC wrapper that makes it easy to embed secure computation p
 
 # Installation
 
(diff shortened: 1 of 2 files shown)


samee:
What are the perl dependencies?

jackdoerner:
MakeMaker and Data-Dumper are two dependencies that are apparently necessary (it won't compile without them installed) that are included as parts of larger packages in Ubuntu. Specifically, MakeMaker is included with perl-modules-5.22, and Data-Dumper with libperl5.22. I can't tell you _exactly_ what they're needed for without further investigation.

samee (approved):
Thanks for the changes.

PR #13, opened 2016-10-11
Added the copy construct to libobliv.a

jackdoerner:
Added the copy construct to libobliv.a and modified makefile to accommodate.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -36,6 +36,7 @@ stamp-h: config.h.in config.status
 	./config.status
 
 CC=@CC@
+CPP=cpp
 @DEFAULT_COMPILER@=1
 
 # We have to use _build because of OCaml's bug #0004502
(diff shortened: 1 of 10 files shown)


samee:
Thanks Jack. Is it also possible to move some of the `copy.oh`-specific tests in here? It'd be nice to have some examples that actually exercise these new functions, however trivial they might be.

jackdoerner:
Sure, I can do that. To be clear, you the ones in the "algorithm" repository, right? Or do you mean some others that I'm not aware of?

samee:
You can just copy these over: https://github.com/samee/sqrtOram/tree/master/test
'algorithms' may also have some. Anything with names testshuffle/testswap/testCopy are good candidates. If it depends on anything other than copy.oh, don't use that.

jackdoerner:
OK, there they are.

samee:
Thanks. Merged. Curious: why did you have to use cpp for dependency generation? Was it causing problems?

jackdoerner:
Yep. Interestingly, gcc refuses to do dependency generation for .oc files (at least on my machines), but cpp works without complaint, even though all the same options are passed. Sqrt ORAM uses cpp for dependency generation, which is where I got the idea from, so I suspect this is a problem you have also solved at least once in the past.

samee:
Huh, go past-me!

PR #22, opened 2017-01-29
Fixes for integer overflows, plus protocol splitting

jackdoerner:
I've changed a couple of things here:

Many int32s enlarged so that we can handle large arrays of obliv data
Network profiling turned on permanently
Added a generic interface for protocol splitting, plus implementations for Yao (sans OT)

I haven't yet finished OT splitting, so feeding inputs doesn't work on threads that have been split from the original. I'm submitting a PR anyway because it will likely be a few weeks before I have time, and I know a couple of projects would find the existing changes useful.

### DIFF
--- a/src/ext/oblivc/copy.c
+++ b/src/ext/oblivc/copy.c
@@ -11,6 +11,8 @@
 #include<stdio.h>
 #define HERE printf("Party %d: %s:%d\n",p,__FILE__,__LINE__)
 
+#define USE_PLAIN_OT
+
 struct OcShareContext
 {
 #ifdef USE_PLAIN_OT
@@ -260,6 +262,30 @@ void ocFromShared_impl(ProtocolDesc* pd,
   // Let's go with malloc for now, we change that later.
   struct OcShareContext* ctx = protoShareCtx(pd);
   YaoProtocolDesc* ypd = pd->extra;
+#ifdef USE_PLAIN_OT
+  OTsender* sender;
+  OTrecver* recver;
+  if (ctx == NULL) {
+    sender = &ypd->sender;
+    recver = &ypd->recver;
+  } else {
+    sender = &ctx->sender;
+    recver = &ctx->recver;
+  }
+#else
+  struct HonestOTExtSender* sender;
+  struct HonestOTExtRecver* recver;
+  if (ctx == NULL) {
+    sender = ypd->sender;
+    recver = ypd->recver;
+  } else {
+    sender = ctx->sender;
+    recver = ctx->recver;
+  }
+#endif
+
+  
+
   int i,j,p = pd->currentParty(pd);
   if(p==1)
   {
@@ -267,11 +293,11 @@ void ocFromShared_impl(ProtocolDesc* pd,
     yao_key_t *key1 = malloc(n*bits*YAO_KEY_BYTES);
 #ifdef USE_PLAIN_OT
     for(i=0;i<n*bits;++i) yaoKeyNewPair(ypd,key0[i],key1[i]);
-    ctx->sender.send(ctx->sender.sender,
+    sender->send(sender->sender,
                      (char*)key0,(char*)key1,n*bits,YAO_KEY_BYTES);
 #else
     struct CorrFunXorArgs a = {.len=YAO_KEY_BYTES,.mask=ypd->R};
-    honestCorrelatedOTExtSend1Of2(ctx->sender,
+    honestCorrelatedOTExtSend1Of2(sender,
         (char*)key0,(char*)key1,n*bits,YAO_KEY_BYTES,corrFunSameXor,&a);
     ypd->icount+=n*bits;
 #endif
@@ -293,9 +319,9 @@ void ocFromShared_impl(ProtocolDesc* pd,
     bool *sel = malloc(n*bits*sizeof(bool));
     unpackBools(sel,n*bits,src);
 #ifdef USE_PLAIN_OT
-    ctx->recver.recv(ctx->recver.recver,(char*)key,sel,n*bits,YAO_KEY_BYTES);
+    recver->recv(recver->recver,(char*)key,sel,n*bits,YAO_KEY_BYTES);
 #else
-    honestCorrelatedOTExtRecv1Of2(ctx->recver,(char*)key,sel,n*bits,
+    honestCorrelatedOTExtRecv1Of2(recver,(char*)key,sel,n*bits,
         YAO_KEY_BYTES);
 #endif
     ypd->icount+=n*bits;
--- a/src/ext/oblivc/obliv.h
+++ b/src/ext/oblivc/obliv.h
@@ -18,6 +18,7 @@ void protocolAddSizeCheck(ProtocolDesc* pd);
 int protocolConnectTcp2P(ProtocolDesc* pd,const char* server,const char* port);
 int protocolAcceptTcp2P(ProtocolDesc* pd,const char* port);
 void cleanupProtocol(ProtocolDesc*);
+void splitProtocol(ProtocolDesc*, ProtocolDesc*);
 
 void setCurrentParty(ProtocolDesc* pd, int party);
 void execDebugProtocol(ProtocolDesc* pd, protocol_run start, void* arg);
@@ -32,5 +33,6 @@ bool execNpProtocol_Bcast1(ProtocolDesc* pd, protocol_run start, void* arg);
 void execNnobProtocol(ProtocolDesc* pd, protocol_run start, void* arg, int numOTs, bool useAltOTExt);
 
 size_t tcp2PBytesSent(ProtocolDesc* pd);
+int tcp2PFlushCount(ProtocolDesc* pd);
 
 #endif // OBLIV_H
--- a/src/ext/oblivc/obliv_bits.c
+++ b/src/ext/oblivc/obliv_bits.c
@@ -886,11 +886,13 @@ void splitYaoProtocolExtra(ProtocolDesc* pdout, ProtocolDesc * pdin) {
   ypdout->extra = NULL;
   ypdout->fixedKeyCipher = ypdin->fixedKeyCipher;
   if (pdout->thisParty == 1) {
-    gcry_randomize(ypdout->R,YAO_KEY_BYTES,GCRY_STRONG_RANDOM);
+    gcry_randomize(&ypdout->gcount,sizeof(ypdout->gcount),GCRY_STRONG_RANDOM);
+    osend(pdin,1,&ypdout->gcount,sizeof(ypdout->gcount));
+    memcpy(ypdout->R,ypdin->R,YAO_KEY_BYTES);
     gcry_randomize(ypdout->I,YAO_KEY_BYTES,GCRY_STRONG_RANDOM);
-    ypdout->R[0] |= 1;   // flipper bit // NOTE: ASSUME POINT AND PERMUTE IS TRUE
     //ypdout->sender = honestOTExtSenderAbstract(honestOTExtSenderNew(pdout,2));
   } else {
+    orecv(pdin,2,&ypdout->gcount,sizeof(ypdout->gcount));
     //ypdout->recver = honestOTExtRecverAbstract(honestOTExtRecverNew(pdout,1));
   }
   pdout->extra=ypdout;
--- a/src/ext/oblivc/obliv_bits.c
+++ b/src/ext/oblivc/obliv_bits.c
@@ -56,7 +56,6 @@ static struct stdioTransport stdioTransport
 void protocolUseStdio(ProtocolDesc* pd)
   { pd->trans = &stdioTransport.cb; }
 
-//#define PROFILE_NETWORK
 // TCP connections for 2-Party protocols. Ignores src/dest parameters
 //   since there is only one remote
 typedef struct tcp2PTransport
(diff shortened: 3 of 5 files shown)


src/ext/oblivc/obliv_bits.c
```diff
@@ -56,7 +56,6 @@ static struct stdioTransport stdioTransport
 void protocolUseStdio(ProtocolDesc* pd)
   { pd->trans = &stdioTransport.cb; }
 
-//#define PROFILE_NETWORK
```
samee, line 59: Why is it important to have network profiling always enabled? I am okay with it being a runtime flag, which is better than the current state IMO (currently libobliv needs to be rebuilt to enable network profiling). I am also okay with the way libc does it: generate two versions of libobliv, with and without profiling, so that oblivcc can choose which to use based on runtime parameters.

jackdoerner: It's not important that it _always_ be enabled, but rebuilding every time you need network profiling is fairly inconvenient (and it seems to be not an uncommon need - I've been repeatedly asked for network figures for various things), and adding a runtime conditional for each gate seems like it wouldn't be significantly more efficient than the single extra addition that permanent enabling incurs. In other words, if it doesn't incur significant overhead (seems like it should be about the same, overhead-wise, as counting gates), is there a reason to disable it? Building two versions of libobliv is definitely better than being required to rebuild libobliv, but it still implies rebuilding of the application binaries, dealing with flags and linking, &c, and I'm not sure what the benefit would be of linking against the one _without_ network profiling.

Please let me know if I've failed to understand something. My assertion that it causes negligible overhead is based on the metric of running programs with it enabled and not noticing any, rather than any rigorous measurement.

jackdoerner:
This extra commit wasn't meant to end up in this PR, but since it's here, if you approve of it, you might as well integrate it. Basically, it allows ocToShared, ocFromShared, ocFeedShared, and ocRevealShared to do something sensible if ocShareInit hasn't been run. This is useful for me, because I'm writing a library function that uses these functions, and I have no way of knowing whether or not ocShareInit has been called by the application already.

src/ext/oblivc/obliv_bits.c
wangxiao1254: The difference is observable as I can remember. But you can do an experiment on the latest version to make sure.

samee: I also seem to remember that. But as Xiao noted, if you can point to gate-per-second numbers on some benchmarks and tell me that it doesn't matter over LAN (or EC2), I'm okay with it.

jackdoerner: Well, I ran the benchmarks, and stand corrected. The difference is (very slightly) noticeable, at least for CPU-constrained tasks. Results are attached. I will try putting all of the profiling stuff behind a runtime option and see what happens. If I'm going to go to the effort, should I also move gate counting behind an option as well? It stands to reason that if the counting the bytes degrades performance, then maybe counting gates does too.

[benchmarks.zip](https://github.com/samee/obliv-c/files/743203/benchmarks.zip)
(for the record, these were recorded between two C4.2xlarge AWS instances; application code was from [ACK](https://bitbucket.org/jackdoerner/absentminded-crypto-kit/commits/eb959053be9d165324d663aced00ee1d9259abef))

samee: Thanks for the tests Jack. I'll take a look later. 

Do not disable gcount: that needs to be unique for each gate for security.

jackdoerner: Ahh, right. I had forgotten it doubles as the counter for the purpose of encryption.

jackdoerner: Runtime flag is (as I feared) actually much worse than simply doing the additions. If you approve it as the right solution, I'll revert the network profiling changes and add another commit to the PR.

src/ext/oblivc/copy.c
```diff
@@ -260,18 +262,42 @@ void ocFromShared_impl(ProtocolDesc* pd,
   // Let's go with malloc for now, we change that later.
   struct OcShareContext* ctx = protoShareCtx(pd);
   YaoProtocolDesc* ypd = pd->extra;
+#ifdef USE_PLAIN_OT
```
samee, line 265: I think we should strive to remove these mystery hash-defines whenever we can (other than maybe profiling/instrumentation, but even then). If strictly necessary, we can turn them into runtime flags where possible. 

In this case, though, I don't even remember why this ifdef exists, and from the looks of it nobody uses it. I say we clean this out, since we are changing things here. I think it is worth making these cases neatly converge. It is okay if you make that a separate pull request, though, and have a TODO here for now.

src/ext/oblivc/copy.c
```diff
@@ -260,18 +262,42 @@ void ocFromShared_impl(ProtocolDesc* pd,
...
+  OTsender* sender;
+  OTrecver* recver;
+  if (ctx == NULL) {
+    sender = &ypd->sender;
+    recver = &ypd->recver;
+  } else {
+    sender = &ctx->sender;
```
samee, line 272: Wait, who initialized ctx?! If ocShareInit didn't run, these are all dangling pointers.

More importantly, why do we want to call this before ocShareInit?

src/ext/oblivc/obliv_bits.c
samee: Try this: how about we have two separate versions of tcp2PTransport ... one profiled and one unprofiled. That way, user can still pick between the two at runtime, but we won't have an actual if-check on a runtime flag. May get us the best of both worlds. What do you think?

src/ext/oblivc/obliv_bits.c
```diff
@@ -899,6 +861,41 @@ void yaoReleaseOt(ProtocolDesc* pd,int me)
   if(me==1) otSenderRelease(&ypd->sender);
   else otRecverRelease(&ypd->recver);
 }
+
+void splitProtocol(ProtocolDesc* pdout, ProtocolDesc * pdin) {
```
samee, line 868: This should be in a separate commit. From now on, we should try to keep each commit a separate topic so that we can safely undo them without breaking other unrelated things.

src/ext/oblivc/obliv_bits.c
```diff
@@ -899,6 +861,41 @@ void yaoReleaseOt(ProtocolDesc* pd,int me)
...
+  pdout->setBitAnd = pdin->setBitAnd;
+  pdout->setBitOr = pdin->setBitOr;
+  pdout->setBitXor = pdin->setBitXor;
+  pdout->setBitNot = pdin->setBitNot;
+  pdout->flipBit = pdin->flipBit;
+  pdout->thisParty = pdin->thisParty;
+  pdout->trans = pdin->trans->split(pdin->trans);
+  pdout->copyextra = pdin->copyextra;
+  pdout->copyextra(pdout, pdin);
+}
```
samee: Thinking aloud here ... do you think it's better to initialize all these fields with the syntax {.error = ..., .partyCount = ... }? This will provide compile errors for missing fields. I'm worried that somebody will add a new field in ProtocolDesc in future without realizing these have to be updated.

Which reminds me ... you added copyextra. Are you sure we don't have existing places that need to be updated?

src/ext/oblivc/obliv_bits.c
```diff
@@ -929,8 +926,12 @@ void setupYaoProtocol(ProtocolDesc* pd,bool halfgates)
   else ypd->recver.recver=NULL;
 
   dhRandomInit();
+
+  //TODO: why is this ECB?
```
samee, line 937: I think I answered this in an email. Either remove or describe briefly why this is necessary.

src/ext/oblivc/obliv_bits.c
```diff
@@ -899,6 +861,41 @@ void yaoReleaseOt(ProtocolDesc* pd,int me)
...
+}
+
+void splitYaoProtocolExtra(ProtocolDesc* pdout, ProtocolDesc * pdin) {
+  YaoProtocolDesc* ypdin = pdin->extra;
+  YaoProtocolDesc* ypdout = malloc(sizeof(YaoProtocolDesc));
+  ypdout->protoType = ypdin->protoType;
+  ypdout->extra = NULL;
+  ypdout->fixedKeyCipher = ypdin->fixedKeyCipher;
+  if (pdout->thisParty == 1) {
+    gcry_randomize(ypdout->R,YAO_KEY_BYTES,GCRY_STRONG_RANDOM);
```
samee: Wait, I thought we weren't resampling R, since this creates a bug.

src/ext/oblivc/obliv_bits.c
```diff
@@ -899,6 +861,41 @@ void yaoReleaseOt(ProtocolDesc* pd,int me)
...
+  pdout->revealOblivBits = pdin->revealOblivBits;
+  pdout->setBitAnd = pdin->setBitAnd;
+  pdout->setBitOr = pdin->setBitOr;
+  pdout->setBitXor = pdin->setBitXor;
+  pdout->setBitNot = pdin->setBitNot;
+  pdout->flipBit = pdin->flipBit;
+  pdout->thisParty = pdin->thisParty;
+  pdout->trans = pdin->trans->split(pdin->trans);
+  pdout->copyextra = pdin->copyextra;
+  pdout->copyextra(pdout, pdin);
```
samee: maybe call it splitextra, given that's what all other parts of code is calling it? Or change everything to copy instead of split.

src/ext/oblivc/obliv_bits.c
```diff
@@ -899,6 +861,41 @@ void yaoReleaseOt(ProtocolDesc* pd,int me)
...
+}
+
+void splitYaoProtocolExtra(ProtocolDesc* pdout, ProtocolDesc * pdin) {
```
samee: Seems like some of this function is duplicated from the usual constructors. Is it possible to refactorize to reduce duplication?

src/ext/oblivc/obliv_bits.c
```diff
@@ -886,11 +886,13 @@ void splitYaoProtocolExtra(ProtocolDesc* pdout, ProtocolDesc * pdin) {
   ypdout->extra = NULL;
   ypdout->fixedKeyCipher = ypdin->fixedKeyCipher;
   if (pdout->thisParty == 1) {
-    gcry_randomize(ypdout->R,YAO_KEY_BYTES,GCRY_STRONG_RANDOM);
+    gcry_randomize(&ypdout->gcount,sizeof(ypdout->gcount),GCRY_STRONG_RANDOM);
```
samee: Ah I see it's fixed here. Maybe squash the two commits before merge, so we won't have it in a broken stage in-between.

And ... if it's still untested, do test it. Feel free to add a simple test to test/oblivc.

src/ext/oblivc/obliv.h
```diff
@@ -18,6 +18,7 @@ void protocolAddSizeCheck(ProtocolDesc* pd);
 int protocolConnectTcp2P(ProtocolDesc* pd,const char* server,const char* port);
 int protocolAcceptTcp2P(ProtocolDesc* pd,const char* port);
 void cleanupProtocol(ProtocolDesc*);
+void splitProtocol(ProtocolDesc*, ProtocolDesc*);
```
samee, line 21: See if you can move this into the splitting commit. If not, a separate commit here will also do.

src/ext/oblivc/obliv.h
```diff
@@ -32,5 +33,6 @@ bool execNpProtocol_Bcast1(ProtocolDesc* pd, protocol_run start, void* arg);
 void execNnobProtocol(ProtocolDesc* pd, protocol_run start, void* arg, int numOTs, bool useAltOTExt);
 
 size_t tcp2PBytesSent(ProtocolDesc* pd);
+int tcp2PFlushCount(ProtocolDesc* pd);
```
samee, line 36: New commit. Even better if commit message has a short 2-3 lines about why this was useful.

samee (changes requested):
Finally went through the changes. Sorry for the delay. Thanks, most of this looks really good. Some comments below ... just because I have a lot of comments doesn't mean I'm mad or anything. Quite the contrary.

src/ext/oblivc/obliv_bits.c
jackdoerner: This is a much better idea than anything I've come up with or heard suggested so far! Definitely the best of both worlds. I'll write it and get back to you.

src/ext/oblivc/obliv_bits.c
jackdoerner: Point taken. I've clearly not yet gotten the hang of working on projects that other people actually use. My git-fu may not be strong enough yet to make it right retroactively, but I'll give it a try (same comment applies for the other "split into a separate commit" comments, needless to say).

src/ext/oblivc/copy.c
jackdoerner: Yes, the idea here was to allow this code to do something sensible in the case that ocShareInit hasn't been called yet (in which case ctx will be null and everything in the else clause will be a dangling pointer). Why might we want that to be the case? I'm using ocFromShared/ocToShared in a library, and I don't know whether the application code has called ocSharedInit(). As designed, it can't be called twice (or so it seems to me, anyway). The alternative is to add a check so that ocSharedInit() becomes a no-op if it's already been run. I'm perfectly OK with that solution if you'd prefer it, although I'm not sure that it's really "proper" to split the protocol in a library if the application hasn't done so. Thoughts?

src/ext/oblivc/obliv_bits.c
jackdoerner: Yes, I think you're right. Functionally identical, but there's a significant advantage to preventing future mistakes. I'm pretty sure I caught all the places that updates were needed (there aren't too many), but if I take your suggestion, that'll ensure it.

jackdoerner:
Thanks for the review. I'll make changes and get back to you. You needn't worry about being interpreted as unhappy; my other life as an artist has long-since taught me how criticism works. Besides, you're pretty much the only person who's ever given me a serious code review, and I appreciate that.

samee:
Regarding your git-fu: make lots and lots of branches, always keep your original branch till you are happy with a new branch that looks right. git commit --amend and git rebase are your friends for things that have not been merged in somewhere public.

jackdoerner:
I'm closing this request for now - expect a new request with the changes you've specified sometime soon.

PR #24, opened 2017-03-29
Floram support

jackdoerner:
Included here: fixes for integer overflows that can occur in large/long-running computations; a slight modification to the makefile behavior (compiling in the old way will still produce the same result); fallback code plus consolidation (as requested) for the ocShare system in copy.c; a new, profiled TCP transport selectable at runtime without recompiling obliv-c; protocol splitting for Yao (with a generic API).

### DIFF
--- a/src/ext/oblivc/copy.c
+++ b/src/ext/oblivc/copy.c
@@ -13,13 +13,8 @@
 
 struct OcShareContext
 {
-#ifdef USE_PLAIN_OT
   OTsender sender;
   OTrecver recver;
-#else
-  struct HonestOTExtSender* sender;
-  struct HonestOTExtRecver* recver;
-#endif
   BCipherRandomGen *gen,*padder;
   size_t padnonce;
 };
@@ -36,7 +31,6 @@ void ocShareInit(ProtocolDesc* pd)
   ctx->gen = newBCipherRandomGen();
   ctx->padder = newBCipherRandomGen();
   ctx->padnonce = 0;
-#ifdef USE_PLAIN_OT
   if(me==1)
   { ctx->sender = honestOTExtSenderAbstract(honestOTExtSenderNew(pd,2));
     ctx->recver = honestOTExtRecverAbstract(honestOTExtRecverNew(pd,2));
@@ -44,15 +38,6 @@ void ocShareInit(ProtocolDesc* pd)
   { ctx->recver = honestOTExtRecverAbstract(honestOTExtRecverNew(pd,1));
     ctx->sender = honestOTExtSenderAbstract(honestOTExtSenderNew(pd,1));
   }
-#else
-  if(me==1)
-  { ctx->sender = honestOTExtSenderNew(pd,2);
-    ctx->recver = honestOTExtRecverNew(pd,2);
-  }else
-  { ctx->recver = honestOTExtRecverNew(pd,1);
-    ctx->sender = honestOTExtSenderNew(pd,1);
-  }
-#endif
   // TODO eventually I should move this to a dedicated
   // That would also allow proper splitting on multithreaded
   ypd->extra = ctx;
@@ -66,13 +51,8 @@ void ocShareCleanup(ProtocolDesc* pd)
 {
   YaoProtocolDesc* ypd = pd->extra;
   struct OcShareContext* ctx = ypd->extra;
-#ifdef USE_PLAIN_OT
   otSenderRelease(&ctx->sender);
   otRecverRelease(&ctx->recver);
-#else
-  honestOTExtSenderRelease(ctx->sender);
-  honestOTExtRecverRelease(ctx->recver);
-#endif
   releaseBCipherRandomGen(ctx->gen);
   releaseBCipherRandomGen(ctx->padder);
   free(ctx);
@@ -183,7 +163,6 @@ void ocShareMuxes(ProtocolDesc* pd,char* z,
   struct OcShareContext* ctx = protoShareCtx(pd);
   int i;
   memcpy(t2,x0,bufsz); memxor(t2,x1,bufsz); // t2 = x0^x1
-#ifdef USE_PLAIN_OT
   randomizeBuffer(ctx->gen,t,bufsz);
   memxor(t2,t,bufsz);
   if(protoCurrentParty(pd)==1)
@@ -193,18 +172,6 @@ void ocShareMuxes(ProtocolDesc* pd,char* z,
   { ctx->recver.recv(ctx->recver.recver,tr,c,n,eltsize);
     ctx->sender.send(ctx->sender.sender,t,t2,n,eltsize);
   }
-#else
-  struct CorrFunXorArgs a = {.len=eltsize,.mask=t2};
-  if(protoCurrentParty(pd)==1)
-  { genrGateToShare(pd,ctx,t,t2,wc,n,eltsize);
-    //honestCorrelatedOTExtSend1Of2(ctx->sender,t,t3,n,eltsize,corrFunXor,&a);
-    honestCorrelatedOTExtRecv1Of2(ctx->recver,tr,c,n,eltsize);
-  }else
-  { evalGateToShare(pd,ctx,tr,wc,n,eltsize);
-    //honestCorrelatedOTExtRecv1Of2(ctx->recver,tr,c,n,eltsize);
-    honestCorrelatedOTExtSend1Of2(ctx->sender,t,t3,n,eltsize,corrFunXor,&a);
-  }
-#endif
   for(i=0;i<n;++i) memmove(z+i*eltsize,(c[i]?x1:x0)+i*eltsize,eltsize);
   memxor(z,tr,bufsz);
   memxor(z,t ,bufsz);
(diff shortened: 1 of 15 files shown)


src/ext/oblivc/copy.c
```diff
@@ -193,18 +172,6 @@ void ocShareMuxes(ProtocolDesc* pd,char* z,
...
-#else
-  struct CorrFunXorArgs a = {.len=eltsize,.mask=t2};
-  if(protoCurrentParty(pd)==1)
-  { genrGateToShare(pd,ctx,t,t2,wc,n,eltsize);
-    //honestCorrelatedOTExtSend1Of2(ctx->sender,t,t3,n,eltsize,corrFunXor,&a);
-    honestCorrelatedOTExtRecv1Of2(ctx->recver,tr,c,n,eltsize);
-  }else
-  { evalGateToShare(pd,ctx,tr,wc,n,eltsize);
-    //honestCorrelatedOTExtRecv1Of2(ctx->recver,tr,c,n,eltsize);
-    honestCorrelatedOTExtSend1Of2(ctx->sender,t,t3,n,eltsize,corrFunXor,&a);
```
samee, line 205: Hmm. This is interesting. Are we losing the ability to do correlated OTs?

src/ext/oblivc/obliv_bits.c
```diff
@@ -26,6 +26,15 @@ static __thread ProtocolDesc *currentProto;
 static inline bool known(const OblivBit* o) { return !o->unknown; }
 
 // --------------------------- Transports -----------------------------------
+
+static bool transportProfilingEnabled = false;
+
+void transportEnableProfiling(bool newVal) {
```
samee: "EnableProfiling" sounds like a verb phrase, which shouldn't take in a bool. Either call it "SetProfiling", or have two separate enable/disable functions with no parameter.

src/ext/oblivc/obliv_bits.c
```diff
@@ -26,6 +26,15 @@ static __thread ProtocolDesc *currentProto;
 static inline bool known(const OblivBit* o) { return !o->unknown; }
 
 // --------------------------- Transports -----------------------------------
+
+static bool transportProfilingEnabled = false;
```
samee: Globals always have weird race condition issues. I know it seems like it shouldn't be a problem here, but then it always bites back when you least expect. Plus it is less granular. I can think of two alternatives that I like better:

1) Move this variable inside ProtocolDesc, so as to have only one ugly global pointer to struct to ever deal with.
2) Even better, have people call different functions. E.g. instead of protocolAcceptTcp2P(), they call protocolAcceptProfiledTcp2().

In either case, we have to worry about copying these fields when copying structs. In option 2, there is an easy choke point where we can do this for all copying, split(). Might be a little harder for ProtocolDesc. What do you think?

src/ext/oblivc/obliv_bits.c
```diff
@@ -67,123 +77,114 @@ typedef struct tcp2PTransport
...
-	/*if(res<=0) { perror("TCP read error: "); return res; }*/
-	/*n2+=res;*/
-  /*}*/
   while(n>n2)
-  { res = fread(n2+(char*)s,1,n-n2, tcpt->sockStream);
-	if(res<0 || feof(tcpt->sockStream)) { perror("TCP read error: "); return res; }
-	n2+=res;
+  { 
+    res = fread(n2+(char*)s,1,n-n2, tcpt->sockStream);
+	  if(res<0 || feof(tcpt->sockStream)) { perror("TCP read error: "); return res; }
```
samee: While we are at it, let's fix these indentations. They were introduced in a tab/space argument :)

src/ext/oblivc/obliv_bits.c
```diff
@@ -364,24 +364,22 @@ void protocolAddSizeCheck(ProtocolDesc* pd)
   t->cb.recv=sizeCheckRecv;
   t->cb.cleanup=sizeCheckCleanup;
 }
-// ---------------------------------------------------------------------------
+
+// --------------------------- Protocols -----------------------------------
```
samee: I'm assuming these are just shuffling things around, and not real code changes. Didn't look too closely.

src/ext/oblivc/obliv_bits.c
```diff
@@ -364,24 +364,22 @@ void protocolAddSizeCheck(ProtocolDesc* pd)
...
+int ocCurrentParty() { return currentProto->currentParty(currentProto); }
+int ocCurrentPartyDefault(ProtocolDesc* pd) { return pd->thisParty; }
+
+ProtocolDesc* ocCurrentProto() { return currentProto; }
+void ocSetCurrentProto(ProtocolDesc* pd) { currentProto=pd; }
 
 void cleanupProtocol(ProtocolDesc* pd)
   { pd->trans->cleanup(pd->trans); }
 
 void setCurrentParty(ProtocolDesc* pd, int party)
```
samee: While at it, why not have this function be near ocSetCurrentProto two lines above :)

samee (changes requested):
Pretty minor convention-related comments. In general, this looks great. Thanks!

samee:
Huh, didn't realize "requesting change" needs a large red X icon.

jackdoerner:
Note: I added a new commit containing protocol splitting. This is the last major change for this PR hopefully. I will go back and address your comments now. As you might expect, using protocols with buffered transports in threads that have dependencies upon one another is a good way to end up with deadlocks. I've tried to mitigate this as much as possible in the relevant parts of the code I'm contributing to obliv-c, but it's more-or-less impossible to prevent a programmer from introducing deadlocks in application code. We may want to add protocol aware locks as an additional mitigation measure, which should help somewhat.

src/ext/oblivc/copy.c
jackdoerner: Yes, this is in response to your previous [comment](https://github.com/samee/obliv-c/pull/22#discussion_r101928250). Perhaps I misinterpreted? The main obliv-c code uses the OTSender and OTRecver constructs, so I kept the branch that was compatible with those. To be honest, I'm not sure what the implications are. Perhaps I should have done something else. If so, advise.

src/ext/oblivc/obliv_bits.c
jackdoerner: The reasoning behind using a global switch was to avoid the machinery associated with propagating profiled-ness through split protocols and to give users a protocol-agnostic way to enable it, assuming we have other protocols in the future. Point taken, though. I will look at making the relevant adjustments. Ironically, I originally wrote this with a protocolAcceptProfiledTcp2() function, but changed it.

src/ext/oblivc/obliv_bits.c
jackdoerner: Indeed

src/ext/oblivc/copy.c
samee: You didn't misinterpret anything. I was just being short-sighted. I _think_ this will double the network cost of muxing xor shares. But it was meant to be a temporary hack anyway. The right solution is probably to demand a new kind of abstraction that provides only correlated OT. Let me think for a bit.

src/ext/oblivc/obliv_bits.c
jackdoerner: I would like to state, for the record, that although I am a proponent of tabs, I'm an even stronger proponent of consistency. Fixed this below.

jackdoerner:
Also: the most critical thing here is that my changes in the protocol-splitting commit are secure. Please pay special attention to those.

src/ext/oblivc/obliv_bits.c
```diff
@@ -307,22 +306,23 @@ static ProtocolTransport* tcp2PSplit(ProtocolTransport* tsrc)
...
   return CAST(tnew);
 }
 
+static ProtocolTransport* tcp2PSplitProfiled(ProtocolTransport* tsrc)
+{
+  tcp2PTransport* t = CAST(tsrc);
+  ((tcp2PTransport*)t)->flushCount+=2;
+  if(!t->isClient) t->bytes+=sizeof(in_port_t);
+  
+  return tcp2PSplit(tsrc);
```
samee: These magic values are likely to silently fall out of sync with tcp2PSplit over time. Can we find a better way? Maybe just add a comment on each sensitive line in tcp2PSplit to remind our future selves that a corresponding change needs to be made in the Profiled version.

jackdoerner: Agreed. We could use the oflush function that I added, which would take care of the profiled case automatically, but we'd have to be sure that no one does a plain fwrite (as opposed to osend). Thoughts on that?

jackdoerner: I've done some refactoring to avoid this. Turned out not to be too troublesome.

src/ext/oblivc/copy.c
samee: Given that right now I don't know of any major users of these xor shares, I am okay with introducing a bandwidth regression here. Just file a bug about this saying that a new COTsender/COTrecver needs to be implemented for efficient shares.

samee (commented):
SplitProto, not Transport. Interesting. Is this for the multithreaded support?

jackdoerner:
ocSplitProto splits the underlying transport when it's called. Yes, it's for multithreading - the idea is you call ocSplitProto and pass the new ProtocolDesc to the child thread.

src/ext/oblivc/copy.c
jackdoerner: OK, that sounds reasonable. I am using the XOR shares for my current project and would probably see at least a slight performance gain. Nothing huge though.

src/ext/oblivc/obliv_bits.c
```diff
@@ -364,25 +384,53 @@ void protocolAddSizeCheck(ProtocolDesc* pd)
...
+    .setBitNot = pdin->setBitNot,
+    .flipBit = pdin->flipBit,
+    .thisParty = pdin->thisParty,
+    .trans = pdin->trans->split(pdin->trans),
+    .splitextra = pdin->splitextra,
+    .cleanextra = pdin->cleanextra,
+    .extra = NULL
+  };
+  if (pdout->splitextra != NULL && pdin->extra != NULL) pdout->splitextra(pdout, pdin);
+  oflush(pdin); oflush(pdout);
```
samee: Do we need to flush both pdin and pdout? More importantly, isn't this step Yao-specific, and should go into splitextra()?

src/ext/oblivc/copy.c
samee: I mean, you are also welcome to try and resolve that bug :) Just doesn't have to be in this pull request.

samee:
Other than the minor split change, I think I am ready to merge this in.

src/ext/oblivc/copy.c
jackdoerner: I'm not really worried about it - just wanted to note that there is at least one user. If I have enough time I might look at it, but I wouldn't bet on that being in the next couple of weeks.

samee:
The interface here is really confusing. I am assuming that if I click "Merge pull requests", it will do the right thing.

jackdoerner:
I think so? To be honest, I've never been on the other end of one of these.

PR #27, opened 2017-04-05
Fixed stdio transports, which were broken by recent updates.

jackdoerner:
Also, I added alternative instructions to the readme for the millionaires' test.

It's worth noting that protocol splitting definitely doesn't work when using stdio. I'm not sure how it would be possible to rectify that. Perhaps it would be best simply to refuse to split (return false or something like that)?

### DIFF
--- a/src/ext/oblivc/obliv_bits.c
+++ b/src/ext/oblivc/obliv_bits.c
@@ -43,16 +43,23 @@ static int stdioSend(ProtocolTransport* pt,int dest,const void* s,size_t n)
 
 static int stdioRecv(ProtocolTransport* pt,int src,void* s,size_t n)
 { 
-  bool *p = stdioFlushFlag(pt);
-  if(*p) { fflush(stdout); *p=false; }
+  pt->flush(pt);
   return fread(s,1,n,stdin); 
 }
 
-static void stdioCleanup(ProtocolTransport* pt) {}
+static int stdioFlush(ProtocolTransport* pt) {
+  bool *p = stdioFlushFlag(pt);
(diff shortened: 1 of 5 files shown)


samee:
> Perhaps it would be best simply to refuse to split (return false or something like that)?

Absolutely. I mean, in theory you could certainly tag each message with a "channel header" or something of the sort and then multiplex on the other side. But it's not worth it in my opinion.

jackdoerner:
Yes, I eventually came to that conclusion myself and put changes to that effect in the second commit. Trying to multiplex data from multiple threads over stdio sounds like asking for nasty corner cases to me.

src/ext/oblivc/obliv_bits.c
```diff
@@ -402,26 +402,32 @@ void setCurrentParty(ProtocolDesc* pd, int party)
...
-    .setBitNot = pdin->setBitNot,
-    .flipBit = pdin->flipBit,
-    .thisParty = pdin->thisParty,
-    .trans = pdin->trans->split(pdin->trans),
-    .splitextra = pdin->splitextra,
-    .cleanextra = pdin->cleanextra,
-    .extra = NULL
-  };
-  if (pdout->splitextra != NULL && pdin->extra != NULL) pdout->splitextra(pdout, pdin);
+bool ocCanSplitProto(ProtocolDesc * pdin) { return (pdin->trans->split != NULL && pdin->splitextra != NULL && pdin->extra != NULL); }
```
samee: Please keep lines to 80 chars, not least so I can actually see the whole text :)

samee (commented):
This looks good. Just do the linebreak and we'll be done.

PR #28, opened 2017-04-15
Added Mac OS compatibility.

jackdoerner:
This simple change seems to fix the compilation problem on Mac OS.
Note: from what I can tell, the src/ext directory is serving double duty for both CIL plugins and libobliv.a's source code, even though these two things are unrelated. In the future, should we think about separating them?

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -36,7 +36,7 @@ stamp-h: config.h.in config.status
 	./config.status
 
 CC=@CC@
-CPP=cpp
+CPP=@CPP@
 @DEFAULT_COMPILER@=1
 
 # We have to use _build because of OCaml's bug #0004502
@@ -58,7 +58,7 @@ CILLIB_TARGETS=
 CILLY_EXE_FILES=
 
 CIL_PLUGINS_DIR = src/ext
-CIL_EXCLUDE_PLUGINS =
+CIL_EXCLUDE_PLUGINS = oblivc
 CIL_PLUGINS = $(addprefix $(OBJDIR)/,$(filter-out $(addprefix $(CIL_PLUGINS_DIR)/, $(CIL_EXCLUDE_PLUGINS)),$(wildcard $(CIL_PLUGINS_DIR)/*)))
 
 CIL_DEFAULT_PLUGINS = $(patsubst $(CIL_PLUGINS_DIR)/%/default,cil.%,$(wildcard $(CIL_PLUGINS_DIR)/*/default))
@@ -128,7 +128,7 @@ $(OCPARTS:%=$(DEPENDDIR)/%.d): $(DEPENDDIR)/%.d: $(OCSRCDIR)/%.c $(DEPENDDIR)
 	echo '	$$(CC) -c $$< -o $$@ -I $$(OCSRCDIR) $$(CFLAGS)' >> $@
 
 $(OOCPARTS:%=$(DEPENDDIR)/%.od): $(DEPENDDIR)/%.od: $(OCSRCDIR)/%.oc $(DEPENDDIR) 
-	$(CPP) -MM $< -I$(OCSRCDIR) -MT $(OBJDIR)/$*.oo > $@
+	$(CPP) -x c -MM $< -I$(OCSRCDIR) -MT $(OBJDIR)/$*.oo > $@
 	echo '	$$(BINDIR)/oblivcc -c $$< -o $$@ -I $$(OCSRCDIR) $$(CFLAGS)' >> $@
 
 # TODO update ocamlutil/Makefile.ocaml no longer used


samee:
Sweet! If anybody asks about Mac, I'll just redirect all questions to you from now on.

But, are you sure this is Mac-compatible? E.g. try running a test that has calloc or memset in it and see if that still works on a Mac. Or just try running hamming distance to start with. I'm asking because I remember there being some Mac-specific built-in macros in standard header files that were not supported by CIL.

jackdoerner:
Happy to receive them

I ran sqrt ORAM on it, and it worked. That said, while this is _necessary_ for Mac compatibility, it's not sufficient. On my system I'm using gcc (the real one, not llvm's cc) from macports, and application code needs to be compiled with some extra flags an environment variables. That's why I haven't changed the readme yet. Until someone figures out a better way, it's probably best not to advertise that it can be made to work, since doing so requires some effort. In the tests I've done, everything seems to work aside from multithreading (and I believe the failures there are due to my application code, not oblivc).

samee:
This patch completely breaks my linux build on a fresh repository, in a way that is not even recoverable with `make clean`. Dependency generation gets botched, so `make` ends with this:
```bash
$ git clone https://github.com/jackdoerner/obliv-c.git fresh
$ cd fresh
$ git log -1
commit 88e373fc03d6e683f727fc13d9e2f3e7ae969ae1
Author: Jack Doerner <j@ckdoerner.net>
Date:   Sat Apr 15 14:47:58 2017 -0400

    Added Mac OS compatibility.
$ ./configure
$ make
[...]
_build/depends/copy.od:1: *** recipe commences before first target.  Stop.
$ cat _build/depends/copy.od 
	$(BINDIR)/oblivcc -c $< -o $@ -I $(OCSRCDIR) $(CFLAGS)
```
If you compare this file with `copy.od` from a working repository, you will see that it is missing the recipe header.
```bash
$ cat $(working_repo)/_build/depends/copy.od
_build/copy.oo: src/ext/oblivc/copy.oc src/ext/oblivc/obliv_common.h \
 src/ext/oblivc/obliv_types.h src/ext/oblivc/obliv_bits.h \
 src/ext/oblivc/obliv_types_internal.h src/ext/oblivc/copy.oh \
 src/ext/oblivc/obliv_copy_internal.h src/ext/oblivc/obliv_yao.h
	$(BINDIR)/oblivcc -c $< -o $@ -I $(OCSRCDIR) $(CFLAGS)
```

jackdoerner:
Funny - that didn't happen when I tried it on my linux machine. Let me find an ubuntu box and try to reproduce. Good job you checked though. I'm guessing that @CPP@ isn't getting defined given the symptom.

Edit: reproduced. Seems @CPP@ picks up a value of gcc -E on at least some distros. There is some difference between the ways gcc -E and cpp behave when passed things that aren't actually valid C code (i.e. obliv-c). I will try to find some resolution.

jackdoerner:
Since no one is using my fork, I simply force-pushed a fix overtop of the existing commit. Breaking git etiquette, I know, but it avoids having commits that won't build in the main obliv-c repo. Try pulling/building again.

What happened:
There is a critical difference between cpp and gcc -E, which is that the former seems to automatically interpret all of its input files as c code unless told otherwise, where as the latter uses file extensions (and possibly other data?) to guess what file type the sources are, and aborts if it can't. On systems where @CPP@ got set to gcc -E, this would cause failures. In the past, both of us worked around this simply by hardcoding cpp (you did this for sqrt ORAM, I did it when I added copy.oc to obliv-c), but that doesn't work when you need to be able to set the preprocessor at compile time, which is necessary on Mac OS. I fixed it by adding a flag to manually specify that .oc files should be interpreted as c code.

samee:
Looks good. Build and tests work. Thanks for the fixes!

> but that doesn't work when you need to be able to set the preprocessor at compile time, which is necessary on Mac OS.

Why is this needed on Mac OS?

> Line 61: `CIL_EXCLUDE_PLUGINS = oblivc`

What kind of error messages were you seeing? Might be worth recording them here for posterity.

> Since no one is using my fork, I simply force-pushed a fix overtop of the existing commit. Breaking git etiquette, I know, ...

Yeah, whatever works. But a better procedure is to have all changes be on topic-branches, so you can always create more branches/delete old ones when this happens. In the end, I will pull from some branch that only has the commits you want. This also gives you an easier undo strategy in case things get screwed up.

jackdoerner:
>> but that doesn't work when you need to be able to set the preprocessor at compile time, which is necessary on Mac OS.

> Why is this needed on Mac OS?

Mac OS ships with LLVM's `cpp` in `/usr/bin`, and you don't want to replace it since both Apple's native build tools and macports depend upon it (not sure if homebrew is similar since I don't use that). Likewise for `/usr/bin/gcc`, which links to LLVM's `cc` by default. Both of these choke on the technically-malformed obliv-c files they are asked to parse during the build process (even if you force them to be parsed as standard C using `-x`). So, to compile obliv-c on Mac OS, one must first install the real `gcc` and its version of `cpp`, then instruct the build process to use these alternates instead of the defaults.

>>   Line 61: `CIL_EXCLUDE_PLUGINS = oblivc`

> What kind of error messages were you seeing? Might be worth recording them here for posterity.

This is needed because the cctools (Mac OS) version of `ar` behaves a bit differently than the GNU binutils version, and unforunately ocaml hardcodes the location of the `ar` binary, so it's not possible to tell the build process to use an alternate, as you can with `cpp`/`gcc`/&c. The error message you get is this one:

```
ocamlfind ocamlopt -a -o src/ext/oblivc.cmxa
+ ocamlfind ocamlopt -a -o src/ext/oblivc.cmxa
ar: no archive members specified
```

i.e. `ar` refuses to create an empty archive (which it's being instructed to do because `src/ext/oblivc` is not a CIL plugin).

Incidentally, there's a chance that obliv-c will now also build on BSD (whence I assume the cctools `ar` probably originates), but I can't guarantee that.

samee:
Thanks for the details!

jackdoerner:
You're welcome. Probably should have put them in the original PR. I will at some point write some detailed Mac OS instructions, but it may be a few weeks before I have time, since I need to start with a fresh install in order to verify that I haven't missed any deps.

PR #30, opened 2017-05-15
Floating Point Support for Obliv-C

darioncassel:
# Purpose

Floating point support has been added to Obliv-C as an "obliv float". 

# Changes Made

Nearly every file related to the Obliv-C CIL extension has been changed. The most important changes were to:
    - [src/ext/processObliv/processObliv.ml](src/ext/processObliv/processObliv.ml)
    - [src/ext/processObliv/oblivUtils.ml](src/ext/processObliv/oblivUtils.ml)
    - [src/cil.ml](src/cil.ml)
    - [src/ext/oblivc/obliv_bits.c](src/ext/oblivc/obliv_bits.c)
    - [src/ext/oblivc/obliv_bits.h](src/ext/oblivc/obliv_bits.h)
    - [src/ext/oblivc/obliv_common.h](src/ext/oblivc/obliv_common.h)
    - [src/ext/oblivc/obliv_bits.c](src/ext/oblivc/obliv_bits.c)
    - [src/ext/oblivc/obliv_types_internal.h](src/ext/oblivc/obliv_types_internal.h)
    - [src/ext/oblivc/obliv_types.h](src/ext/oblivc/obliv_types.h)
    - [src/ext/oblivc/obliv.h](src/ext/oblivc/obliv.h)

# Documentation

Comments have been added throughout the additions where appropriate to explain the changes. A guide to extending the language has been added to the README.

# Testing

The tests/testFloat folder contains tests for each standard function used with obliv floats (plus, minus, multiplication, etc). The tests/linreg folder contains an implementation of linear regression performed with the obliv float data type.


samee (commented):
Thanks a lot Darion. It's a lot of code in one pull, so it might be a few weeks before I have reviewed it all. But I'm sure it will be useful in the end.

jackdoerner:
Hello there. Unlike Samee, I don't have a proper job, so I've gone ahead and had a preliminary look at things. Also unlike Samee, I am not the BDFL of this project, so my opinions don't actually mean anything unless he happens to agree with them. That said, I am a daily user of and occasional contributor to Obliv-c, for whatever that's worth. Forewarning: I intend to be quite harsh in my critique, but that doesn't mean I don't think this is good work. Although I personally have never had a need for floating point in obliv-c, it does seem to me to be one of the most significant missing pieces on the path to being a fully general tool for MPC. Thanks for taking it on, and good work! AFACT, this PR only covers single-precision, but once you're done with the feedback/merge cycle, it seems like double precision shouldn't be too much additional trouble, and it would be well worthwhile.

First, some general notes, not associated with any specific piece of code:

With respect to the overall approach, the idea of having functions implemented as large, opaque circuits makes me a bit uncomfortable, since it essentially precludes any maintenance or future understanding without reference to the source (which I understand is tinygarble, in this case?), and even then makes things difficult. That said, I don't expect any efficient circuit-based IEEE floating point implementation to be simple or pretty, and I appreciate the work you've done to modularize it such that the circuits can be replaced fairly easily: if we're going to have big, opaque circuits, then they should be isolated as much as possible. Sometime in the indeterminate future, it would be very nice to have circuits that can handle arbitrary-precision FP calculations, similar to the way the integer circuits work now, but I understand that this is 1) a much bigger task, and 2) probably incompatible with the approach of adapting circuits from tinygarble.

So far as I can tell, your code is meant to be compatible with any underlying protocol, and the `floatProtocol` you added is for the purpose of debugging the new circuits. Let me know if this is a wrong interpretation. If it's not, I think it's best to roll whatever float protocol functionality is useful to users into the main debug protocol; if this isn't possible (you would know better than I), then I would suggest that it be renamed to 'debugFloatProtocol' or similar, and `ifdef`ed out by default. My reasoning is this: it took me some time to determine with certainty that floating point works in yao, and the float protocol is in fact a debugging tool. I assume it would be even more difficult a determination for anyone who isn't familiar with the codebase. Best to avoid confusion, especially if the functionality won't be needed by anyone who isn't working on obliv-c itself.

Finally, some minor general housecleaning notes:
- there's some build detritus in here that needs to be removed. A couple of `*.o` files, some `a.out` files, and a `.DS_store` (which is an apple metadata file), at least.
- the SCDtoObliv tool should go in some sort of `tools` (or similar) directory, rather than being in the root of the project.
- the `test/testFloat` and `test/linreg` directories should be subdirectories of `test/oblivc` (e.g. `test/oblivc/float` and `test/oblivc/linreg` respectively)
- The float circuits should probably go in a subdirectory, since there's a bunch of them and they're closely related. The main src directory is already getting crowded.
- IMO, notes on extending obliv-c should go in a dedicated document, and should _not_ be in the README (although perhaps the secondary document should be referenced by the README), because users should be solving problems be writing code _in_ the language rather than by extending it if at all possible. Otherwise, thanks for writing documentation!

Those are my general notes so far; over the next day or two, I will go over the files and leave inline comments on specific things. I may also add more here, or edit what I've already written as my understanding improves. Again, thanks for doing this! As my opinions are unimportant, you should not feel obligated to reply to anything I say unless you deem it worthwhile.

src/cil.ml
```diff
@@ -6746,7 +6751,7 @@ let rec xform_switch_stmt s break_dest cont_dest = begin
        * label_break: ; // break_stmt
        *
        * The default case, if present, must be used only if *all*
-       * non-default cases fail [ISO/IEC 9899:1999, �6.8.4.2, �5]. As
+       * non-default cases fail [ISO/IEC 9899:1999, �6.8.4.2, �5]. As
```
jackdoerner: My browser doesn't seem to have any glyphs for these characters, whereas the old ones were fine. Why the change?

src/cil.mli
```diff
@@ -2549,6 +2549,7 @@ val unoblivType : typ -> typ
 
 val isOblivFunc : typ -> bool
 val isOblivInt : typ -> bool
+val isOblivFloat : typ -> bool (*CHANGE*)
```
jackdoerner: Perhaps you could document (even if only here) what this should be changed to, and why.

src/ext/oblivc/obliv_bits.h
```diff
@@ -40,13 +43,17 @@ void __obliv_c__setBitOr(OblivBit* dest,const OblivBit* a,const OblivBit* b);
 void __obliv_c__setBitXor(OblivBit* dest,const OblivBit* a,const OblivBit* b);
 void __obliv_c__setBitNot(OblivBit* dest,const OblivBit* a);
 void __obliv_c__flipBit(OblivBit* dest); // Avoids a struct copy
+void __obliv_c__floatAdd(OblivBit* dest,const OblivBit* a,const OblivBit* b);
```
jackdoerner: The naming convention here seems to be different than elsewhere. Is there a reason?

src/ext/oblivc/obliv_transport.c
```diff
@@ -0,0 +1,343 @@
...
+
+static void stdioCleanup(ProtocolTransport* pt) {}
+
+// Extremely simple, no multiplexing: two parties, one connection
+static struct stdioTransport stdioTransport 
+  = {{2, NULL, stdioSend, stdioRecv, stdioCleanup},false};
+
+void protocolUseStdio(ProtocolDesc* pd)
+  { pd->trans = &stdioTransport.cb; }
+
```
jackdoerner: This PR seems to overwrite a whole bunch of commits I made, which changed the way network profiling worked (specifically, PROFILE_NETWORK was removed, which is why I noticed). I suspect other commits by myself and others are also overwritten, though I haven't specifically looked for obsolete code. Github does not detect the conflict for some reason - possibly because of the fact that you split everything into separate files (which is a very good thing, otherwise).

src/ext/oblivc/obliv_bits.c
```diff
@@ -1842,6 +681,7 @@ void feedOblivInputs(OblivInputs* spec, size_t count, int party)
     { __obliv_c__##ot rv; \
       OblivInputs spec; \
       setupObliv##tname(&spec,&rv,v); \
+      spec.src_f = v; \
```
jackdoerner: Given that `spec.src_f` is unioned with `spec.src`, won't this overwrite the src variable with bogus data for all types that _aren't_ float? If I am understanding correctly, this should be done in `setupOblivFloat()` to avoid that problem.

src/ext/processObliv/processObliv.ml
```diff
@@ -49,6 +49,9 @@ let oblivIntTarget = ref dummyType
 let oblivShortTarget = ref dummyType
 let oblivLongTarget = ref dummyType
 let oblivLLongTarget = ref dummyType
+let oblivFloatTarget = ref dummyType (*CHANGE*)
```
jackdoerner: As elsewhere, these (*CHANGE*) markers should probably be annotated with the reason.

src/ext/processObliv/processObliv.ml
```diff
@@ -594,6 +634,12 @@ let rec codegenUncondInstr (instr:instr) : instr = match instr with
         | BNot -> setUnop "__obliv_c__setBitwiseNot" v e loc
         | LNot -> setUnop "__obliv_c__setLogicalNot" v e loc
         end
+    | TFloat(kind, a) when hasOblivAttr a ->
+        begin match op with
+        | Neg -> setUnop "__obliv_c__setNegF" v e loc
+        | BNot -> E.s (E.error "Bitwise operation on float invalid!\n")
```
jackdoerner: As a style note: the existing errors are not terminated with an exclamation point or a newline (though maybe the newline has a reason?). I would also suggest more specificity, to make it easier for programmers to figure out the problem. Perhaps "bitwise not operator unsupported for obliv-float type". Even better, go find out what error GCC produces and make it read similar to that one if you can (unless their error isn't very good). This comment applies elsewhere in this file as well.

test/linreg/genData.py
```diff
@@ -0,0 +1,37 @@
+# Samuel Havron
+# https://www.github.com/samuelhavron
+# MIT License
```
jackdoerner: Although I will grant you that the MIT license is essentially identical to BSD, it probably makes sense to License all of Obliv-C the same way.

test/oblivc/aes/aes.c
```diff
@@ -17,16 +17,17 @@ int main(int argc,char* argv[])
     if(argc<2) fprintf(stderr,"Port number missing\n");
     else if(argc<3) fprintf(stderr,"Party missing\n");
     else fprintf(stderr,"string missing\n");
-    fprintf(stderr,"Usage: %s <port> <1|2> <string>\n",argv[0]);
-    fprintf(stderr,"  Party 1 provides key in hex\n");
-    fprintf(stderr,"  Party 2 provides plaintext in hex\n");
+    fprintf(stderr,"Usage: %s <port> <--|remote_host> <string>\n",argv[0]);
```
jackdoerner, line 20: This is much better than the original way I think (no more recompiling to change the IP Address!), but I still feel like there's a lot of redundancy here among the various tests (which will only get worse as more are added). This isn't something I think you need to fix before merge (or that you should feel obligated to fix at all) - just something to note for the future. I have a partial solution to this problem in one of my repos, which may or may not be appropriate to adapt.

src/ext/processObliv/processObliv.ml
darioncassel: Apologies! The original purpose of those markers was for me to keep track of changes across commits. I thought I removed all of them but clearly I overlooked a few. I can definitely fix this for the PR.

src/cil.ml
darioncassel: I'm not too sure what happened here...perhaps my text editor messed something up? I can switch those back to the old characters.

src/cil.mli
darioncassel: That line actually did not "change to" its current form; it was freshly added. In this case I think a comment explaining the change isn't quite necessary as it is a simple line. I will remove the comment. (See my remark below on the other CHANGE comment).

src/ext/oblivc/obliv_bits.h
darioncassel: The reason that the naming convention is different is because the floatAdd function represents an arithmetic circuit-based operation whereas the other functions are just individual bitwise gates.

src/ext/processObliv/processObliv.ml
darioncassel: The newline is a mistake and should be removed. The reason for the error the way it is is because we want to discourage the use of bitwise operations on floats in general, not just obliv floats. GCC does not produce an error; I believe it is just undefined behavior and may surface a warning.

test/linreg/genData.py
darioncassel: That license was set by Sam Havron. I have no problem with changing it but we should probably ask Sam first if he is ok with it.

src/ext/oblivc/obliv_transport.c
darioncassel: I'm not sure what happen here...could it be that I had starting making my changes on a branch that did not have your changes? And now I'm merging back to master so it might not appear, or actually be, overwritten? Could you point me to where I can see this overwritten code? Thanks, and sorry about that!

src/ext/oblivc/obliv_bits.h
jackdoerner: The ones above it yes, but elsewhere we have at least these three styles:
floatAdd
setFloatKnown
setPlainAddF

Unless I am mistaken, all of these represent circuits over many bits. Is there a reason for them to be differently named (an honest question - maybe there is)?

src/ext/oblivc/obliv_transport.c
jackdoerner: > could it be that I had starting making my changes on a branch that did not have your changes

Yes, but git is supposed to detect conflicts when you merge. I suspect that's failed because a lot of the code is now in different files than it started in. I will get back to you on specific code segments, but for now I would think that most changes to obliv_bits.c since commit 0b3dc3d might end up being overwritten.

src/ext/processObliv/processObliv.ml
jackdoerner: >  GCC does not produce an error; I believe it is just undefined behavior

Are you sure? I just tried it and got `Error: integralPromotion: not expecting float` if I do it inside an obliv-c source file, which may not be the best error message of all time, but it gets the point across. If I do it inside a regular .c file and compile with gcc, I get `error: invalid operands to binary << (have 'float' and 'int')`, which is a pretty good error I think. I'm not sure what the standard says (or fails to say), but it seems the compilers won't allow this behavior regardless.

> we want to discourage the use of bitwise operations on floats in general

While I understand the sentiment here, I personally believe that a highly specific error message has a huge value in the context of debugging, which outweighs any 'general discouragement' a less specific variant may imply. Especially since it seems that both CIL and gcc forbid the shifting of floats anyway (unless I am mistaken).

darioncassel: Alright, I see what you mean; I'll update the error message to be more specific.

src/ext/oblivc/obliv_bits.h
darioncassel: Yes, while those three all do work over more than one bit, the naming conventions for the second and third are actually copied from their int-addressing counterparts so that they fit in with the rest of the code.

src/ext/oblivc/obliv_bits.c
darioncassel: Although it may seem that way, that is not the behavior I observe. It's been a while since I wrote that segment but I believe that the assignment to ```src_f``` is harmless exactly because it is unioned with ```src```; I am just "placing data" in a bucket that can either be interpreted as a float or as an int; nothing is being overwritten (and the ```setupObliv##tname``` functions, which all call ```__obliv_c__setupOblivBits```, were going to do that anyways). In any case, in my testing I have not encountered any issues with this.

jackdoerner: In memory, floats and ints are stored in entirely different formats (as I'm sure you know, since you did just write an entire codebase dealing with float-formatted data). Since the type of `spec.src_f` is float, I believe it is the case that `v` will be implicitly converted to a float here (remember, `v` is integral in all but one instantiation of this function), and then the memory (or at least the first 32 bits of it) underlying the union of `spec.src_f` and `spec.src` will be overwritten with float-formatted data, regardless of the input type. There is definitely an overwrite happening, because you're performing an assignment. The _value_ stored in `spec.src` may be the same, but the _representation_ will be different, and this is important.  Your float debugging protocol always reads `spec.src_f`, and therefore will print some value that looks correct (plus/minus representational errors that happen when converting to float) regardless of whether the input was an int or a float. The debug protocol, on the other hand, reads `spec.src` (see `dbgProtoFeedOblivInputs`). Because `spec.src` is unioned with `spec.src_f`, no type conversion will be done during this read - rather, the raw float-formatted bits will be interpreted as an integer (which is wrong). Have you tried the debug protocol? Regardless of whether there is some confounding factor here that makes it somehow work out correctly, my interpretation is that `spec.src` is meant to hold the incoming data in its original form, which means that it should _not_ be converted to or stored as a float unless it was one to begin with.

Maybe I have missed something here - if so, please let me know. For the record, I haven't actually run the debug protocol with your changes yet, but I can, if it helps.

darioncassel: Thanks for that detailed response. Again, I apologize as it's been a while since I wrote that code but I am fairly sure that this actually does work. The float debugging protocol is not what I used for testing -- the tests are in the ```test/``` folder. Specifically there is ```test/testFloat``` that tests all of the floating point operations and there is ```test/linreg``` that runs linear regression. Both of these use Yao's protocol, not the float debugging protocol. In order to make sure that it does work, I just ran some additional code (with Yao's protocol) that uses a mix of ```obliv int```s and ```obliv float```s and it executed without issue. Again, I don't recall why I wrote it this way but I will look into it to find the reason.

jackdoerner:
An additional note: I tried to compile the last version referenced in this PR, and received the following error for several lines in `obliv_floatProto.c`:

`error: 'OblivBit {aka struct OblivBit}' has no member named 'floatValue'`

`obliv_floatProto.c` is included in `dualex.c`, although it wasn't immediately obvious to me why this would be the case, since `dualex.c` deals with yao protocols. Deleting the include seems to allow it to compile.

I checked manually, and the latest version in your repo does indeed seem to have a discrepancy between `OblivBit` as defined and as referred to. Since you mentioned that it compiles for you, are you sure you've pushed all your changes? Or have I perhaps missed some instruction?

src/ext/oblivc/obliv_bits.c
jackdoerner: No worries. These things go slowly. For my part, I seem to have been confused by the existence of a similarly named float-support repo, which I mistook for the one you're attempting to PR (thus the now-deleted comment about failed compilation about which you may have received some sort of notification).

Having switched to the correct repo, I added the following lines near the top of `testFloatOps.oc` (at line 15):

```
int ref = ocBroadcastInt(rand(), 1);
obliv int x2 = feedOblivInt(ref, 2);
int testout;
revealOblivInt(&testout, x2, 0);
printf("DEBUG: %d, %d\n", ref, testout);
```

When I ran it, both parties printed `DEBUG: 16807, 1183010304`

Thoughts?

darioncassel: Ok, I see what you mean :/

In my testing I was not calling feedOblivInt and instead just doing a direct assignment. I forgot that that would not work because that calls setup without going through the feed function. I really apologize for being stubborn and for the messiness! You are correct, the following fix seems to work (moving the ```spec.src_f = v``` to the setup function):
```
void setupOblivFloat(OblivInputs* spec, __obliv_c__float* dest, float v)
  { 
    __obliv_c__setupOblivBits(spec,dest->bits,v,__bitsize(v));
    spec->src_f = v;
  }
```

I will include this as part of the other improvements I am doing to this pull request unless you see any problems with that.

Thanks :)

jackdoerner: Haha, not a problem. The resolution you propose seems fine to me, but @samee is ultimately the judge. Some day we need some thorough regression tests to help with this sort of thing.

test/linreg/genData.py
havron: Yes, you're quite welcome to change that to be flush with the rest of the repo's BSD license!

evansuva:
Is anything happening with this?

samee:
It's a lot of code 🙂

I have started going through them, so expect at least a preliminary set of comments by this weekend.

samee:
Since it has already been a long time since the pull request, I am skipping the usual back and forth and doing as much of the fixes as I can by myself.

Ok, so I think I finished rebasing. I pushed it out to a separate branch `float-support` for now. Darion, can you please check that I didn't break anything? I'll merge it in if everything looks good.

First of all, Jack's "housecleaning" notes on his first comment are all spot on. Please do not include binaries, or `.DS_Store`, or other development files you don't think will be useful for other developers of Obliv-C. This also includes other TinyGarble repositories you added in temporarily. This includes intermediate commits. You can either `git rebase -i` to edit those commits and remove those files, or even better, just add those to `.gitignore`.

Generally, think of commits as a unit of change that we might want to rollback later. If a bug is found in future, we don't want to land on an intermediate commit that doesn't build. It is okay for you to have them in your private branch, but do squash those out before a pull request.

Always declare functions before using them, even if compiler doesn't complain. E.g. `obliv_float_*_circuit` (now in a new header file `obliv_float_ops.h`). It allows better typechecking.

The README was clearly meant for your own repository, so I am removing that.

Below is a list of changes I have made, with some additional comments. If any of my changes breaks things, let me know. Other than that, please resolve these if you can.

### `SCDtoObliv/SCDtoObliv.py`

```diff
+SCDtoObliv.py - Can read, evaluate, and output
+    garbled circuits formatted for Obliv-C or
+    for the pyOblivTest program.
+
```

Users of Obliv-C may not know what an SCD file is. Adding a link.

```diff
+class util_cppReadOp:
```

Is anybody using this class, or should this be removed?

### `bin/oblivcc`
```diff
-CILLYARGS=""
+CILLYARGS=
```
Not sure what's happening here. Reverting.

### `src/cil.ml`
```diff
+  (* TODO: add something here for obliv floats *)
```
Can you quickly test what actually happens here? I don't want to merge this if we are silently returning the wrong size in sizeof(obliv float). If it's just 32-bits of `OblivBit`s, I *think* you can just change the "isOblivSimple t" guard condition to accept "isOblivSimple t || isOblivFloat t". The rest of the code should just work.


### `src/cil.mli`
```diff
+val isOblivFloat : typ -> bool (*CHANGE*)
```
I don't see what needs to change here.


### `src/ext/oblivc/dualex.c`
```diff
+#include "obliv_floatProto.c"
```
I don't see where this has been used. Reverting.


### `src/ext/oblivc/obliv.h`
```diff
+void execFloatProtocol(ProtocolDesc* pd, protocol_run start, void* arg);
```
Looks like a temporary test function of some sort, one that nobody is using right now. Is this useful for future development, or should I remove it?
```diff
+uint64_t yaoGateCount(void);
+
```
Not sure how this got added here. Reverting.

### `src/ext/oblivc/obliv_bits.c`
```diff
-  OblivBit* dest=vdest;
-  while(size-->0)
-  { __obliv_c__assignBitKnown(dest,value&1);
-    value>>=1; dest++;
-  }
+    OblivBit* dest=vdest;
+    while(size-- > 0) {
+        __obliv_c__assignBitKnown(dest, value & 1);
+        value >>= 1;
+        dest++;
+    }
```
If you really want to do cleanups on the public repository, keep them in a separate pull request. This is just cluttering up diffs.

```diff
+// carryIn and/or carryOut can be NULL, in which case they are ignored
+void __obliv_c__setBitsAddF (void* vdest,void* carryOut
+                           ,const void* vop1,const void* vop2
+                           ,const void* carryIn
+                           ,size_t size)
+{
+  OblivBit *dest=vdest;
+  const OblivBit *op1=vop1, *op2=vop2;
+  obliv_float_add_circuit(dest, op1, op2);
+}
+
 ```
`carryIn` and `carryOut` are *always* being ignored here. Collapsing to `setPlainAddF()`.
```diff
+void __obliv_c__condNegF (const void* vcond, void* vdest
+                        ,const void* vsrc, size_t n)
+{
+  OblivBit *dest=vdest;
+  const OblivBit *op1=vsrc, *op2=vcond;
+  obliv_float_neg_circuit(dest, op1, op2);
+}
+
```
Again, this doesn't actually look at `vcond`. For integers I had a `condNeg()` because there were useful in implementing signed division: I used it to negate a number only if a certain condition was true. It doesn't look like you plan to do that here. I collapsed it to `setNegF()` for now, but feel free to introduce a conditional version if needed. It's very easy to do.

In `__obliv_c__setEqualToF`:
```diff
+  __obliv_c__assignBitKnown(dest,0);
```
why is this line necessary? Same question for other comparisons. I'm guessing you have it here because it was crashing on manual tests due to these being uninitialized. But in C code generated from Obliv-C, `obliv bool`s are always initialized to zero anyway.

```diff
+      // Use the "union trick" to convert wd to a proper float
+      // since wd is stored as a widest_t
+      union ItoF {
+        widest_t i;
+        float f;
+      };
+      union ItoF conv;
+      conv.i = wd;
+      *dest = conv.f;
```
I think you could also replace all of this with `*dest = *(float*)(&wd);`. What we really want is to use some sort of an endian-neutral wrapper, but that's for a later fix.


### `src/ext/oblivc/obliv_bits.h`
```diff
+void __obliv_c__floatAdd(OblivBit* dest,const OblivBit* a,const OblivBit* b);
```
Which function is this? Looks like an orphan. Removing.


### `src/ext/oblivc/obliv_floatProto.c`

Do you want to keep this file? I'd like to remove it, unless you think it will be useful in future development.

### `src/ext/oblivc/obliv_types.h`
```diff
 struct OblivInputs {
-  unsigned long long src;
+  union {
+    unsigned long long src;
+    float src_f;
+  };
   struct OblivBit* dest;
   size_t size;
```
Same here: if execFloatProto() gets removed, this should be reverted as well.


### `src/ext/processObliv/processObliv.ml`

```diff
-| TFloat(_,a) -> if hasOblivAttr a then Some "unimplemented" else None
+| TFloat(_,a) -> None
```
We still need to raise an error on 'obliv double' and 'obliv long double'. Fixed.
```diff
+let oblivFloatTarget = ref dummyType (*CHANGE*)
+let oblivDoubleTarget = ref dummyType (*CHANGE*)
+let oblivLongDoubleTarget = ref dummyType (*CHANGE*)
```
Change to what?
```diff
+        if (isint st && isint t) || (isfloat st && isfloat st)
```
Typo:) Fixing.

```ocaml
| Set(dv,CastE(dt,Lval sv),loc) when isOblivInt dt ->
    begin match unrollType dt,unrollType (typeOfLval sv) with
    | TInt(dk,da), TInt(sk,sa) when hasOblivAttr sa ->
        if isSigned sk then
          setIntExtend "__obliv_c__setSignExtend" dv dk sv sk loc
        else setIntExtend "__obliv_c__setZeroExtend" dv dk sv sk loc
    | _ -> instr
    end
```
Are you sure you don't need a `TFloat` case here? E.g. what happens when you do `obliv float x = my_int;` Or vice-versa? Or, you may also need to support conversions such as `obliv float x = 5.0l;`.

```diff
 let cmpLtFuncs = ("__obliv_c__setLessThanUnsigned"
                  ,"__obliv_c__setLessThanSigned")
+let cmpLtFuncsF = ("__obliv_c__setLessThanF"
+                  ,"__obliv_c__setLessThanF")
 let cmpLeFuncs = ("__obliv_c__setLessOrEqualUnsigned"
                  ,"__obliv_c__setLessOrEqualSigned")
+let cmpLeFuncsF = ("__obliv_c__setLessThanEqF"
+                  ,"__obliv_c__setLessThanEqF")
```
This looks convoluted. Signedness does not apply to floats, they are always signed. Directly using `setComparison`.


### `src/frontc/cabs2cil.ml`
```diff
+
+    | TFloat _, TPtr _ -> result
+    | TPtr _, TFloat _ -> result
```
This looks out of place. Why is this necessary?


### `test/testFloat/`

Will anything break if I move this to `test/oblivc/testFloat/` ?


### `test/testFloat/dbg.h`

Looks like a duplicate file. Please include from a single place.

samee:
Also, please resolve Jack's comments as well, they look unresolved as well. You can make commits and link to them here, I'll pull and squash them in.

jackdoerner:
A note: @samee's branch _appears_ to fix the clobbering of commits from the main branch, which I noted previously, but it also appears to revert the splitting of obliv_bits. I am in favor of splitting obliv_bits again, in a later commit, and I'm happy to contribute when the time comes.

A second note: @samee's branch doesn't build for me. On both Fedora 26 with GCC 6.4.1 and Mac OS 10.12.6 with GCC 5.4.0 via macports, I get the following error:
`src/ext/oblivc/obliv_bits.c: In function '__obliv_c__setNegF':
src/ext/oblivc/obliv_bits.c:1753:34: error: 'vcond' undeclared (first use in this function)
   const OblivBit *op1=vsrc, *op2=vcond;`

I haven't tried to debug it yet. I have some deadlines soon, so I won't have much time in the next few days, but after that, if this PR hasn't been resolved yet, I'll take another look.

samee:
You are right, build broken. Fixed that particular error. Old tests still broken due to missing headers.

samee:
I *think* I fixed the old tests. Need to run them more thoroughly.

Regarding splitting files: that was intentional. But I agree with Jack, let's do them separately. We have enough going on on this thread.

jackdoerner:
Can confirm that it now builds, and I can build the ACK library with it, but unfortunately, that library's tests fail almost (but not quite) uniformly, whereas they pass with obliv-c's master branch. For reference, this includes known-good codebases such as sqrt ORAM and circuit ORAM, in addition to loads of my own code.

samee:
Hmm. We should seriously invest in at least a checklist of tests to run before merging anything here. That checklist can then easily become a shell script.

Confirming sqrtOram build failure with these changes. For future debugging, here are the steps for testing:

    git clone git@github.com:samee/sqrtOram.git
    cd sqrtOram
    export OBLIVC_PATH=path/to/oblivc
    make build/bench/benchsqrtAccess
    build/bench/benchSqrtAccess -- 1234 100 10 &
    build/bench/benchSqrtAccess localhost 1234 100 10

Segfaults with current float-support branch of Obliv-C, but succeeds on main HEAD on obliv-c branch. I'll have to debug later, unless somebody else fixes it.

jackdoerner:
I agree with this wholeheartedly. Ideally there should eventually be a proper test suite, but for now, a couple of projects in which we have confidence would be much better than nothing. Sqrt ORAM and one of the AES circuits (in addition to the demo one there is [this one](https://bitbucket.org/jackdoerner/absentminded-crypto-kit/src/bb457cae153d0324374f30e64a43f09a246d75e8/src/oaes.oc?at=master&fileviewer=file-view-default), which is much smaller) would probably be a good place to start. There are finicky bits related to protocol management, XOR-sharing, &c that aren't covered by those cases, but those features are less widely used, so it's probably less catastrophic if they get broken.

Maybe a good place to start with a more complete solution would be a C validation suite? Most of them seem to be for-profit, but [GCC includes a free one](https://gcc.gnu.org/onlinedocs/gccint/C-Tests.html). Regardless, this is a bit off topic here.

darioncassel:
Thanks for reviewing my code, Samee. I'll be looking into your comments over the next few days and integrating the appropriate fixes. Once I'm done, I'll rebase all of the relevant commits into a single one. 

As for the segfault with sqrtOram, I believe this might be caused by an issue that Jack pointed out earlier, namely "Given that spec.src_f is unioned with spec.src, won't this overwrite the src variable with bogus data for all types that aren't float? If I am understanding correctly, this should be done in setupOblivFloat() to avoid that problem." I have the fix done locally, but not committed yet. I will ensure that it solves the problem and push it with the rest of my changes.

Also with regards to the "(\*CHANGE\*)" comments, these were actually supposed to be removed but some of them seemed to have slipped through. They don't indicate changes that are required but rather changes that have been made in adding float support. I'll remove them and add in proper comments as appropriate.

darioncassel:
@samee I've gone over the changes you've made and I believe that I've resolved the issues that were brought up either by you or by Jack on this thread. I've cherry-picked the latests commits from your float-support branch in order to absorb your changes into my PR, and on top of those I've made a few commits that fix the other issues. Please let me know how you want this rebased / modified before it can be merged into the main repo as I've tried rebasing it but it doesn't seem to come out cleanly. I'll be trying rebasing again after you get a chance to review these latest changes.

## Responses to the points you brought up:

>  ### SCDtoObliv/SCDtoObliv.py
> `+class util_cppReadOp:`
> Is anybody using this class, or should this be removed?

You're right, that class is no longer necessary so I have removed it.

> ### src/cil.ml
> `+  (* TODO: add something here for obliv floats *)`
> Can you quickly test what actually happens here? I don't want to merge this if we are silently returning the wrong size in sizeof(obliv float). If it's just 32-bits of OblivBits, I think you can just change the "isOblivSimple t" guard condition to accept "isOblivSimple t || isOblivFloat t". The rest of the code should just work.

I believe that you're correct about the required change. I have changed the guard `isOblivSimple t` to `isOblivSimple t || isOblivFloat t` since it is just 32 OblivBits throughout.

> ### src/ext/oblivc/obliv.h
> `+void execFloatProtocol(ProtocolDesc* pd, protocol_run start, void* arg);`
> Looks like a temporary test function of some sort, one that nobody is using right now. Is this useful for future development, or should I remove it?

That function is for my `FloatProtocol` which I was using for testing during the development of the floating point support. I found it useful then but now you're right in saying that it doesn't have that much value. You can remove it and the `FloatProtocol` file if you'd like.

> ### In __obliv_c__setEqualToF:
> `+  __obliv_c__assignBitKnown(dest,0);`
> why is this line necessary? Same question for other comparisons. I'm guessing you have it here because it was crashing on manual tests due to these being uninitialized. But in C code generated from Obliv-C, obliv bools are always initialized to zero anyway.

I'm not sure why it's necessary but the reason that I have it is because all of the comparisons (the ones that were there before I started adding floating point support) have it. For example `__obliv_c__setNotEqual` has `__obliv_c__assignBitKnown(dest,0);`. Should these lines be removed globally or am I missing something?

> ```
> +      // Use the "union trick" to convert wd to a proper float
> +      // since wd is stored as a widest_t
> +      union ItoF {
> +        widest_t i;
> +        float f;
> +      };
> +      union ItoF conv;
> +      conv.i = wd;
> +      *dest = conv.f;
> ```
> I think you could also replace all of this with *dest = *(float*)(&wd);. What we really want is to use some sort of an endian-neutral wrapper, but that's for a later fix.

I tried out that replacement and it doesn't seem to work. I actually tried out a few things before resorting to using the "union trick" but all of them did not work. What are you thoughts on this?

> ```
> | Set(dv,CastE(dt,Lval sv),loc) when isOblivInt dt ->
>     begin match unrollType dt,unrollType (typeOfLval sv) with
>     | TInt(dk,da), TInt(sk,sa) when hasOblivAttr sa ->
>         if isSigned sk then
>           setIntExtend "__obliv_c__setSignExtend" dv dk sv sk loc
>         else setIntExtend "__obliv_c__setZeroExtend" dv dk sv sk loc
>     | _ -> instr
>     end
> ```
> Are you sure you don't need a TFloat case here? E.g. what happens when you do obliv float x = my_int; Or vice-versa? Or, you may also need to support conversions such as obliv float x = 5.0l;.

I'm not sure -- are we trying to support that kind of behavior? I can add in support for that but I hadn't because I didn't realize that we would want those kinds of conversions.

> ### src/frontc/cabs2cil.ml
> ```
> +
> +    | TFloat _, TPtr _ -> result
> +    | TPtr _, TFloat _ -> result
> ```
> This looks out of place. Why is this necessary?

You're right, that has now been removed.

> ### test/testFloat/
> 
> Will anything break if I move this to test/oblivc/testFloat/ ?

Things would have broken because of relative paths in the Makefile but I've fixed that and moved both` testFloat` and the `linreg` test to `test/oblivc` (and changed their Makefiles to make use of the common Makefile in `test/oblivc/common`).

> ### test/testFloat/dbg.h
> 
> Looks like a duplicate file. Please include from a single place.

I've now included that file from a single place.

## Other Changes
### src/ext/processObliv/processObliv.ml
While I was testing out these changes at first compilation was not terminating on my test cases. I determined the cause to be the removal of `| TFloat(k,_) -> fst fnames` from 
```
let setComparisonUS fnames dest s1 s2 loc = 
  let optype = typeOfLval s1 in
  let fname = match unrollType optype with
  | TInt(k,_) -> if isSigned k then snd fnames else fst fnames
  | _ -> E.s (E.error "Cannot operate on obliv values of type %a" d_type optype)
  in
  setComparison fname dest s1 s2 loc
```
which was done during the changes in your rebase. Adding back that case allows compilation to terminate and the test cases to run successfully. What was the reason for the removal?

### src/ext/oblivc/obliv_bits.c
Moved `spec.src_f = v` from ` feedOblivInputs` to the setup function:
```
void setupOblivFloat(OblivInputs* spec, __obliv_c__float* dest, float v)
  { 
    __obliv_c__setupOblivBits(spec,dest->bits,v,__bitsize(v));
    spec->src_f = v;
  }
```
As I mentioned earlier, I believe that this was the cause of issues when using the changes in this PR with other code. After this change, running:
```
git clone git@github.com:samee/sqrtOram.git
cd sqrtOram
export OBLIVC_PATH=path/to/oblivc
make build/bench/benchSqrtAccess
build/bench/benchSqrtAccess -- 1234 100 10 &
build/bench/benchSqrtAccess localhost 1234 100 10
```
no longer causes a segfault.

### Misc.
The `(*CHANGE*)` markers have been removed. Most of them were next to self-explanatory one-liners that were usually declarations so comments seemed redundant.

samee:
I can't really tell which commits you want me to cherry pick on top of `float-support` branch. I tried commit `b790d67: Parity with samee/obliv-c branch float-support`, but it still has a lot of diffs with all of my branches here. Can you please list out a few commits that, when played back on `float-support`, nothing is broken?

E.g.:

    git fetch git@github.com:samee/obliv-c.git float-support
    git checkout -b samee-float-support FETCH_HEAD

Now you should have two branches, `fp-support` and `samee-float-support`. Try to play whatever commits you want on top of my branch, fix things, and push it out for me to see.

Let me know if this works.

darioncassel:
@samee I apologize for not being clear about that; what I meant is that I have already cherry-picked the last three commits from your float-support branch and integrated them with my branch. I was thinking we could do a merge-rebase of my branch into your repo. Would that work for you? (The reason that my commits won't play back properly on your branch is that my fp-support branch does not have a lot of the commits you included in float-support. I tried to be minimal with what I was including as part of the PR and thus did not include all of those commits. There should not be a difference/problem when it comes to merging though).

samee:
Merging creates conflicts as well. The conflicts might be simple to solve, though. In any case, if I understand your intentions, you want me to include the following commits from your fp-support branch:

```
$ git log
commit e13886715bb24638f9d9f42bea766ecca40c2b5c
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:11:49 2017 -0400

    Move linreg test to oblivc tests folder, move dbg.h to test/oblivc/common

commit 62acdd3e53e91f0a38e55ead885116a08bbeffde
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:03:34 2017 -0400

    Update testFloat with clearer test explanation messages

commit c7d8bed6f1ebe8282cd2998333c2bb81fbc1463f
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:00:04 2017 -0400

    Fix compilation-breaking changes in processObliv, move testFloat to oblivc tests folder

commit e39e7886224755302814ed7a773fa4e5ba382ec4
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Mon Sep 4 17:08:18 2017 -0400

    Add Makefile changes from samee/obliv-c branch float-support

commit a04058390eec5da6300cfd5d3ee670417132faef
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Mon Sep 4 16:56:47 2017 -0400

    Remove extraneous change markers, improve float op error messages, change genData license

commit b790d6744fc79efffc3114f87890dc6b6210226a
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Mon Sep 4 16:48:44 2017 -0400

    Parity with samee/obliv-c branch float-support

```

Is this what I should review now?

darioncassel:
@samee Actually, of those commits only the top three are needed. The bottom three were just to make my branch functionally equivalent to yours. So these three can be reviewed:

```
commit e13886715bb24638f9d9f42bea766ecca40c2b5c
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:11:49 2017 -0400

    Move linreg test to oblivc tests folder, move dbg.h to test/oblivc/common

commit 62acdd3e53e91f0a38e55ead885116a08bbeffde
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:03:34 2017 -0400

    Update testFloat with clearer test explanation messages

commit c7d8bed6f1ebe8282cd2998333c2bb81fbc1463f
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 5 12:00:04 2017 -0400

    Fix compilation-breaking changes in processObliv, move testFloat to oblivc tests folder
```

I should note that there were a few changes in this commit:
```
commit a04058390eec5da6300cfd5d3ee670417132faef
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Mon Sep 4 16:56:47 2017 -0400

    Remove extraneous change markers, improve float op error messages, change genData license
```
but as is shown in the commit message, these are fairly straightforward changes based on pervious discussion. You might want to approve the error messages I've added though.

samee:
Again, Darion, please clean up diffs for commit before mailing them out, either here, or on some other open source project, or even in a company. In particular:

  * Make the new series of commits from HEAD so you know they don't conflict. Itis okay if the HEAD moves since you started coding, just rebase them off. Test that merging works before emailing them out.
  * Make each commit an easy to understand logically consistent change. If you make them too small, you risk leaving the code in a broken state between commits. If you make them too large, too many unrelated changes show up together, making them harder to understand.

While every project is a little different, here is a reasonable guide to get you started:
https://people.redhat.com/rjones/how-to-supply-code-to-open-source-projects/

For example, I have split up your commit `a0405839` into separate smaller ones.

These things affect future maintainability of a project, specially when most of the code is no longer being contributed by a single developer.

When cherry-picking `c7d8bed`, I saw your comment about removal of `| TFloat(k,_) -> fst fnames` causing problems. Can you give me an example of this, possibly with the input files? I could be wrong, but I don't see any place where the float-related code is calling that function. `setComparisonUS` is for the unsigned/signed case, when we need two separate comparison functions based on the signed-ness of operands. `setComparison`, on the otherhand, is when a single function is enough. Since floats are always signed, we only ever need one function. This is why I removed that line in `bc13208cf`:

```diff
$ git diff bc13208cf{^,} src/ext/processObliv/processObliv.ml
[...]
@@ -501,7 +498,6 @@ let setComparisonUS fnames dest s1 s2 loc =
   let optype = typeOfLval s1 in
   let fname = match unrollType optype with
   | TInt(k,_) -> if isSigned k then snd fnames else fst fnames
-  | TFloat(k,_) -> fst fnames
   | _ -> E.s (E.error "Cannot operate on obliv values of type %a" d_type optype)
   in
   setComparison fname dest s1 s2 loc
[...]
@@ -685,10 +681,10 @@ let rec codegenUncondInstr (instr:instr) : instr = match instr with
         | Shiftlt-> E.s (E.error "Shift operation on float invalid!\n")
         | Ne -> setComparison "__obliv_c__setNotEqualF" v e1 e2 loc
         | Eq -> setComparison "__obliv_c__setEqualToF"  v e1 e2 loc
-        | Lt -> setComparisonUS cmpLtFuncsF v e1 e2 loc
-        | Gt -> setComparisonUS cmpLtFuncsF v e2 e1 loc
-        | Le -> setComparisonUS cmpLeFuncsF v e1 e2 loc
-        | Ge -> setComparisonUS cmpLeFuncsF v e2 e1 loc
+        | Lt -> setComparison "__obliv_c__setLessThanF" v e1 e2 loc
+        | Gt -> setComparison "__obliv_c__setLessThanF" v e2 e1 loc
+        | Le -> setComparison "__obliv_c__setLessThanEqF" v e1 e2 loc
+        | Ge -> setComparison "__obliv_c__setLessThanEqF" v e2 e1 loc
```
Float-related code shouldn't need to use that function at all. So I'm a little surprised that line caused problems, since I don't think it is being invoked when processing floats. Maybe there's something I missed, but I'd like to know the underlying cause of the problem.

darioncassel:
@samee Thanks for reviewing my changes. I apologize for the messiness; I did not realize that you wanted to add my changes back to your float-support branch (I was under the impression that we would merge my fp-support branch) and as such I was not looking at the diffs against your branch. For any future changes I will just start at the HEAD of your branch as you suggest so that changes are easier to integrate. Sorry about that.

With regards to the issue with:
```
-  | TFloat(k,_) -> fst fnames
```
The reason that I came to think it was an issue was that my floating point test suite, `testFloat` (now found in `test/oblivc/testFloat`) was not compiling with that line missing. The error I receive is:
```
Error: Cannot operate on obliv values of type float obliv __attribute__((__simplifyTemp__,
__implicitCast__))
```
which of course is the fall-through branch of the match statement that the removed line belongs to. Now that I consider it again though, this could just mean that I have some malformed code in my [testFloatOps.oc](src/test/oblivc/testFloat/testFloatOps.oc) file, however at the moment I don't see the issue. Let me know what you think.

samee:
No problem.

Regarding those `TFloat` cases, yeah let's not try to solve a problem we can't reproduce or understand. We can remove those pieces.

So for the rest of https://github.com/darioncassel/obliv-c/commit/c7d8bed6f1ebe8282cd2998333c2bb81fbc1463f, it seems you added a case for casts to float, but this doesn't seem to be solving everything. E.g., I tried making the following changes on top of https://github.com/darioncassel/obliv-c/commit/e13886715bb24638f9d9f42bea766ecca40c2b5c:

```diff
$ git diff
diff --git a/test/oblivc/testFloat/testFloatOps.oc b/test/oblivc/testFloat/testFloatOps.oc
index 9ec8a407..f09a0efd 100644
--- a/test/oblivc/testFloat/testFloatOps.oc
+++ b/test/oblivc/testFloat/testFloatOps.oc
@@ -37,6 +37,14 @@ void floatOps(void* args)
     printf("Y_r: %f\n", y_r);
     printf("\n");

+    // Casts
+    obliv int i = obliv_x;
+    int i_r;
+    revealOblivInt(&i_r, i, 0);
+    printf("Casting to int test: expecting 1\n");
+    printf("Casting to int: %d\n", i_r);
+    printf("\n");
+
     // Addition
     obliv_z = obliv_x + obliv_y;
     revealOblivFloat(&z, obliv_z, 0);

$ make
../../..//bin/oblivcc -DREMOTE_HOST=localhost -O3 -I . testFloatOps.oc testFloatOps.c ../common/util.c
testFloatOps.oc: In function ‘floatOps’:
testFloatOps.oc:41:3: error: conversion to non-scalar type requested
     obliv int i = obliv_x;
   ^
```

We want to follow normal C semantics as closely as possible to avoid surprises. In C, we can convert a `float` to `int` either implicitly or by explicit casting (whether or not I agree with the implicit casting behavior is a different story). So you may want to add a truncation circuit so this can be done in their obliv counterparts as well, at least if the user wants a conversion.

Or alternatively, you can leave it open as a bug for now, but at least have a clearer error message that says the feature is unimplemented.

samee:
Btw, I like the way you keep input variables separate in `load_data` of `testFloatOps.c`. I think it is less confusing than the way I tend to do it with both parties using the same variable. It may actually help those unfamiliar with Obliv-C.

darioncassel:
Ok, I agree with what you said. To make sure we're on the same page, going forward are these the things that will be done?

1. Keep the `| TFloat(k,_) -> fst fnames` case (since it handles casts to floats)
2. Add an error message for the case `obliv int <= obliv float`
3. Open an issue for the implementation of a truncation circuit for that case.

If you're in agreement, I will go ahead and add the improved error message and then open an issue for the truncation circuit.

samee:
> 2. Add an error message for the case obliv int <= obliv float
> 3. Open an issue for the implementation of a truncation circuit for that case.

Yes, that sounds quite reasonable. Please make sure we cover all combinations of float/int, float/float/ int/float, int/int combinations, even if it just says "unimplemented". We don't want an undefined behavior.

> 1. Keep the `| TFloat(k,_) -> fst fnames` case (since it handles casts to floats)

No, please remove it. My understanding if your previous message is that we cannot explain the purpose of those lines, in which case I don't want it in the code.

Let me know if you have more questions.

darioncassel:
@samee I've been investigating the `| TFloat(k,_) -> fst fnames` case in `setComparisonUS` and have found out why it was "necessary" earlier, and the fix for this issue. 

The reason why it seemed necessary was actually because of a bug in the logic of the `Set(v,UnOp(op,Lval e,t),loc)` and `Set(v,BinOp(op,Lval e1,Lval e2,t),loc)` branches of `codegenUncondInstr`. Namely, when I was writing the relevant additions necessary to support the obliv float I did not consider the fact that some operators change the expression type `t` to `_Bool obliv` and since a `bool` is just an `int`, this is actually captured by the `TInt(kind,a) when hasOblivAttr a` branch. Thus, obliv float comparisons were falling into the `setComparisonUS` function, and this was actually was not an issue of a cast (not) occurring for the case of `obliv float => obliv int` as we had thought.

The fix for this that I've implemented in
```
commit 3ae1c4110ac4a004b6070194822af4acb1b36b6c
Author: Darion Cassel <darion.cassel@gmail.com>
Date:   Tue Sep 19 20:06:00 2017 -0400

    Modify Set(v,UnOp.. and Set(v,BinOp.. to properly handle operator of type obliv float -> obliv bool
```
is to add an additional check to the `TInt(kind,a) when hasOblivAttr a` branches to look for cases where the operator(s) are actually obliv floats. (An error of "unimplemented..." will be raised in the case that only one of the operators is of the type of `obliv float` in the `BinOp` case.) This approach seems clumsy to me, so if you have an idea for a better way of doing this that'd be great.

A consequence of this fix is that I discovered that the `obliv_bits` header actually had two errors:
1. Incorrect capitalization of `__obliv_c__setLessThanF`
2. Missing declaration for `__obliv_c__setEqualToF`

Those fixes are found in the two commits previous to `3ae1c41`.

My intention was to apply this commits on top of your `float-support` branch but unfortunately, attempting to merge and rebase your branch on top of mine caused a lot of problems so I wasn't able to go through with it... luckily these changes are small and localized enough to be copy & pasted if you'd like or if you'd prefer I can try again rebasing my branch again.

samee:
Wow, I do appreciate you being thorough, but you are probably working too hard here! I was going through your code in https://github.com/darioncassel/obliv-c/commit/3ae1c4110ac4a004b6070194822af4acb1b36b6c, and you have added *lots* of code that are unreachable and unnecessary :) Let me explain.

> I did not consider the fact that some operators change the expression type t to _Bool obliv and since a bool is just an int, this is actually captured by the TInt(kind,a) when hasOblivAttr a branch. 

First of all, great catch! Even I didn't notice it, and I should have. But this conversion is rare. The way CIL works is that it simplifies any C code into a sort of assembly-language-like version of simplified C, before Obliv-C even touches it. Here's a useful experiment you can do. Let us, for instance, try an input mixing ints with floats:

```C
float z = 5.0f + 10;
```
You can see what CIL does to it using this command (put the line above in a function):

    $ path/to/obliv-c/bin/cilly mytest.c -c --save-temps --dosimplify

CIL's simplify plugin will translate it into this (along with other preprocessing cruft we don't care about):
```C
  float z ;
  float __cil_tmp2 ;

  __cil_tmp2 = (float )10;
  z = 5.0f + __cil_tmp2;
```

And this is normal C code, with no hint of `obliv` anywhere. The important part here is that, by the time `processObliv.ml` gets involved, there is no binary expression that deals with mixed operands. The arithmetic addition operator (`PlusA`) will either have only ints or only floats, never both.

So I don't think line 671 of https://github.com/darioncassel/obliv-c/commit/3ae1c4110ac4a004b6070194822af4acb1b36b6c will ever be executed. I tried a few tests with your `testFloat.oc` as well, and I think this is true for most binary ops you have here; you can remove these lines. You'll only need the comparisons, where two floats can produce a bool, since destination type in a comparison differs from operand type. As such, I would also change the pattern-matches a bit to make sure it only matches with destination type bool. So we could change these lines of your commit above (648-652 in `processObliv.ml`):

```ocaml
    | TInt(kind,a) when hasOblivAttr a ->
        begin match unrollType (typeOf (Lval e1)) with
        | TFloat(kind,a) when hasOblivAttr a ->
          begin match unrollType (typeOf (Lval e2)) with
          | TFloat(kind,a) when hasOblivAttr a ->
```
into
```ocaml
    | TInt(IBool, a) when hasOblivAttr a &&
                          isOblivFloat (typeOf (Lval e1)) &&
                          isOblivFloat (typeOf (Lval e2))  ->
       (* handle comparisons between floats *)
    | TInt(kind,a) when hasOblivAttr a ->
       (* leave obliv int binary operations unchanged *)
```

Similarly, bitwise operations on floats: CIL already complains about it long before it gets to your code. You don't need those cases. The situation is similar for unary operators: you have a new `BNot` case that can be removed. Binary NOT doesn't make sense on float, and earlier stages will already catch that. The only one that CIL won't already catch is logical NOT. You can leave that as is (raising the error) or just implement it with a one-line C function that will call `__obliv_c__setNotEqualF` and compare with zero.

Let me know if these make sense, or if I missed some case. You can play around with `testFloatOps.oc` to check my claims. But if you agree, I'll add these changes on.

darioncassel:
Thanks for going through that Samee! I agree with your analysis of the situation; those cases I added do seem unnecessary given the transformations you showed that CIL performs. I think that the simplification you have listed:
```ocaml
| TInt(IBool, a) when hasOblivAttr a &&
                          isOblivFloat (typeOf (Lval e1)) &&
                          isOblivFloat (typeOf (Lval e2))  ->
       (* handle comparisons between floats *)
| TInt(kind,a) when hasOblivAttr a ->
       (* leave obliv int binary operations unchanged *)
```
is valid. Please let me know if there is anything else you need me to do.

jackdoerner:
I found a bug in samee's float-support branch. See line obliv_bits.c, line 2069. This belongs in `setupOblivFloat` rather than `feedOblivFun`. This was previously discussed in my code review of Darion's original PR (somewhere above), but somehow got missed by Samee's pass through the repo. I can confirm that moving this line to `setupOblivFloat` causes libACK's ORAM tests, incl. sqrt, circuit, and floram to pass (on my mac). This test suite is reasonably comprehensive - In addition to initialization tests, each ORAM gets a full pass of sequential writes and reads, and also 2n random reads and writes (where n is the size of the ORAM) of random values. On the other hand, I cannot confirm that floating point ops actually work.

samee:
Sounds fair. See if this fixes it.

jackdoerner:
Can confirm that it now works!

samee:
Merge pushed. Obliv-C now has 'obliv float'.

Thanks both of you for all the effort, specially Darion for implementing the whole thing! Closing out.

evansuva:
Excellent! Thanks to all of you for your hard work on this. I much appreciate Samee's efforts to preserve quality and consistency in the code, all of Jack's contributions to make this work and go through the PRs, and Darion's work on this and patience in seeing things through.

PR #34, opened 2017-06-06
Added convenience functions for revealing and broadcasting arrays

jackdoerner:
To match similar functions for feeding, copying, and sharing/unsharing arrays. None of this is necessary, but I find it obviates a lot of common boilerplate code for me, and I believe the use case is fairly common for other people too.

 It occurs to me that we have two naming conventions here. Maybe in the future, we should choose one?

### DIFF
--- a/src/ext/oblivc/obliv.oh
+++ b/src/ext/oblivc/obliv.oh
@@ -49,14 +49,35 @@ bool revealOblivInt(int* dest, obliv int src,int party);
 bool revealOblivShort(short* dest, obliv short src,int party);
 bool revealOblivLong(long* dest, obliv long src,int party);
 bool revealOblivLLong(long long* dest, obliv long long src,int party);
+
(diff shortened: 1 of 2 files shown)


samee:
Things I might change: on most protocols today, it will be fewer round trips to implement reveal to party 0 (broadcast) as a reveal to party 1 followed by party 2.

jackdoerner:
It's been a while, but IIRC my code didn't touch anything to do with parties? Or are you just suggesting this as an upgrade?

samee:
Optimization, yes. Sometimes you'll see me writing cryptic things like if (party != 1) reveal all to 2; if (party != 2) reveal all to 1;

samee:
Hey Jack, ran into a bunch of unrelated issues when trying to pull this. More details in #49. Long story short, I pushed out the result in branch [ocaml-update](https://github.com/samee/obliv-c/tree/ocaml-update). But I did zero testing. If you have some time to test it out before I do, that would be awesome.

Thanks!

PR #37, opened 2017-07-31
Add cast to ensure C++ compatibility

schoppmp:
The explicit cast allows `obliv.h` to be included from C++ source files, without throwing any errors or warnings. Otherwise, GCC needs to be called with `-fpermissive` to allow implicit casts from void pointers.

### DIFF
--- a/src/ext/oblivc/obliv_types.h
+++ b/src/ext/oblivc/obliv_types.h
@@ -33,7 +33,7 @@ typedef struct OblivBit OblivBit;
 
 typedef char yao_key_t[YAO_KEY_BYTES];
 
-#define NNOB_KEY_BYTES 10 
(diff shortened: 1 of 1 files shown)


samee:
Thanks!
Won't you also need an `extern "C" { ... }` wrapper? I feel like this might give you a linker error if you try to use it from C++.

schoppmp:
Yes, I need to wrap the entire include directive:
```C++
extern "C" {
    #include "obliv.h"
}
```
However, I still need to do the explicit cast (see, for example, this [stackoverflow answer](https://stackoverflow.com/a/40787471))

samee:
Merging for now, but I'll probably add the extern braces later as well.

PR #39, opened 2017-08-24
Resolve naming conflict with libgcrypt 1.8.0

schoppmp:
Feel free to change the name of the function, I just used the first thing that made sense to me. It might also be a good idea to rename `gcry_mpi_point_neg`. It's not defined by libgcrypt (yet), but it also uses its prefix.

### DIFF
--- a/src/ext/oblivc/ot.c
+++ b/src/ext/oblivc/ot.c
@@ -236,7 +236,7 @@ static void npotSend_roundRecvKey(NpotSender* s,NpotSenderState* q)
   { q->PK0 = dhRecv(s->pd,s->destParty); }
 
 // x,y,z are scratch
-static void gcry_mpi_point_copy(gcry_mpi_point_t w, gcry_mpi_point_t v,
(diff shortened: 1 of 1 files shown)


samee:
Awesome! Thanks for the `gcry_mpi_point_neg` suggestion, I might do that. As for the name conflict, do you think it makes sense to just use the new libgcrypt function instead of defining our own? I mean, if it's doing almost the same thing ...

samee:
Little confused about my own past actions: why did I take EC points, negate-then-add, ... instead of just using subtract functions? Were the subtractions not available?!

samee:
Just tried using subtract functions:
```
Fatal: _gcry_mpi_ec_sub_points: Weierstrass not yet supported
```

I might have to patch libgcrypt upstream. But till then, we are doing this the old-fashioned way.

jackdoerner:
As a note on this: while I'm generally in favor of not using shims if we can help it, I have found it to be difficult to get up-to-date versions of libgcrypt on some distributions, so it may behoove us not to depend on recent API additions if we can help it.

PR #42, opened 2017-09-13
Improvements to protocol splitting

jackdoerner:
Instead of generating new OT extensions, all yao protocols share a single extension box, and use GCC atomics to distribute nonces among themselves without collisions. This saves quite a bit of time (400ms on my machines). Updated the protocol splitting test to allow for testing and benchmarking of this change.

### DIFF
--- a/src/ext/oblivc/obliv_bits.h
+++ b/src/ext/oblivc/obliv_bits.h
@@ -2,6 +2,13 @@
 #define OBLIV_BITS_H
 #define __oblivious_c
 
+/* 
+   Note to maintainers: this header should *never* be included in any other
+   header; it will automatically be included in obliv-c source by the compiler,
+   and it contains defines that should never appear in plain C source (whereas
+   including most other obliv-c headers in regular C source is OK)
+   */
+
 //void* memset(void* s, int c, size_t n); // Hack, had to declare memset
 #include<string.h> // memset to zero
 #include<stdbool.h>
--- a/src/ext/oblivc/obliv_bits.h
+++ b/src/ext/oblivc/obliv_bits.h
@@ -5,8 +5,7 @@
 /* 
    Note to maintainers: this header should *never* be included in any other
    header; it will automatically be included in obliv-c source by the compiler,
-   and it contains defines that should never appear in plain C source (whereas
-   including most other obliv-c headers in regular C source is OK)
+   and it contains defines that should never appear in plain C source.
    */
 
 //void* memset(void* s, int c, size_t n); // Hack, had to declare memset
--- a/src/ext/oblivc/obliv_common.h
+++ b/src/ext/oblivc/obliv_common.h
@@ -74,7 +74,7 @@ void npotRecv1Of2(struct NpotRecver* r,char* dest,const bool* sel,int n,int len,
     int batchsize);
 
 struct HonestOTExtRecver* honestOTExtRecverNew(ProtocolDesc* pd,int srcparty);
-struct HonestOTExtRecver* honestOTExtRecverSplit(struct HonestOTExtRecver* t, ProtocolDesc* pd);
+struct HonestOTExtRecver* honestOTExtRecverSplit(ProtocolDesc* pd,struct HonestOTExtRecver* r_src);
 void honestOTExtRecverRelease(struct HonestOTExtRecver* recver);
 void honestOTExtRecv1Of2(struct HonestOTExtRecver* r,char* dest,const bool* sel,
     int n,int len);
@@ -87,7 +87,7 @@ void honestOTExtRecv1Of2Chunk(void* vargs,char* dest,int nchunk,
     int len,bool isCorr);
 
 struct HonestOTExtSender* honestOTExtSenderNew(ProtocolDesc* pd,int destparty);
-struct HonestOTExtSender* honestOTExtSenderSplit(struct HonestOTExtSender* t, ProtocolDesc* pd);
+struct HonestOTExtSender* honestOTExtSenderSplit(ProtocolDesc* pd,struct HonestOTExtSender* s_src);
 void honestOTExtSenderRelease(struct HonestOTExtSender* sender);
 void honestOTExtSend1Of2(struct HonestOTExtSender* s,
     const char* opt0,const char* opt1,int n,int len);
--- a/src/ext/oblivc/obliv_common.h
+++ b/src/ext/oblivc/obliv_common.h
@@ -74,6 +74,7 @@ void npotRecv1Of2(struct NpotRecver* r,char* dest,const bool* sel,int n,int len,
     int batchsize);
 
 struct HonestOTExtRecver* honestOTExtRecverNew(ProtocolDesc* pd,int srcparty);
+struct HonestOTExtRecver* honestOTExtRecverSplit(struct HonestOTExtRecver* t, ProtocolDesc* pd);
 void honestOTExtRecverRelease(struct HonestOTExtRecver* recver);
 void honestOTExtRecv1Of2(struct HonestOTExtRecver* r,char* dest,const bool* sel,
     int n,int len);
@@ -86,6 +87,7 @@ void honestOTExtRecv1Of2Chunk(void* vargs,char* dest,int nchunk,
     int len,bool isCorr);
 
 struct HonestOTExtSender* honestOTExtSenderNew(ProtocolDesc* pd,int destparty);
+struct HonestOTExtSender* honestOTExtSenderSplit(struct HonestOTExtSender* t, ProtocolDesc* pd);
 void honestOTExtSenderRelease(struct HonestOTExtSender* sender);
 void honestOTExtSend1Of2(struct HonestOTExtSender* s,
     const char* opt0,const char* opt1,int n,int len);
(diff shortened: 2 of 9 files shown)


src/ext/oblivc/obliv_common.h
```diff
@@ -86,6 +87,7 @@ void honestOTExtRecv1Of2Chunk(void* vargs,char* dest,int nchunk,
     int len,bool isCorr);
 
 struct HonestOTExtSender* honestOTExtSenderNew(ProtocolDesc* pd,int destparty);
+struct HonestOTExtSender* honestOTExtSenderSplit(struct HonestOTExtSender* t, ProtocolDesc* pd);
```
samee: SenderNew in the line above uses (pd, dest), SenderSplit uses (sender_stuff, pd). How about reordering arguments here? And let's start adding comments on function headers, single line comments are enough for now.
Finally, I think the rest of the file uses `s` for sender, and `r` for receiver. `t` is kind of awkward here. How about `s_src` or `s_parent` or something like that?

I know I don't set a great example wrt API consistency, but let's try to improve on that:)

src/ext/oblivc/obliv_bits.h
```diff
@@ -2,6 +2,13 @@
 #define OBLIV_BITS_H
 #define __oblivious_c
 
+/* 
+   Note to maintainers: this header should *never* be included in any other
+   header; it will automatically be included in obliv-c source by the compiler,
+   and it contains defines that should never appear in plain C source (whereas
+   including most other obliv-c headers in regular C source is OK)
```
samee: Let's remove the part about most other headers. Yes it might be true today, but I don't want to encourage that.

src/ext/oblivc/ot.c
```diff
@@ -677,6 +677,31 @@ senderExtensionBoxNew (ProtocolDesc* pd, int destParty, int keyBytes)
 
   return s;
 }
+SenderExtensionBox*
+senderExtensionBoxSplit (ProtocolDesc * pd, SenderExtensionBox* t)
```
samee: Should `SenderExtensionBox` be `const`? I guess we are changing state of t->keyblock[i] by reading from it, but C won't complain about const-ness since those are referenced through a pointer.

src/ext/oblivc/ot.c
```diff
@@ -717,6 +742,28 @@ recverExtensionBoxNew (ProtocolDesc* pd, int srcParty, int keyBytes)
...
+  char seed[OT_SEEDLEN];
+  int i;
+  r->keyblock0 = malloc(sizeof(BCipherRandomGen*[k]));
+  r->keyblock1 = malloc(sizeof(BCipherRandomGen*[k]));
+  for(i=0;i<k;++i)
+  { randomizeBuffer(t->keyblock0[i], seed, OT_SEEDLEN);
+    r->keyblock0[i] = newBCipherRandomGenByKey(seed);
+    randomizeBuffer(t->keyblock1[i], seed, OT_SEEDLEN);
+    r->keyblock1[i] = newBCipherRandomGenByKey(seed);
+  }
```
samee: Oh ... wait this is actually a more significant change than I thought. Yeah, I'm not so sure this particular method is secure. However, it may be possible to use a variation of this that works. Instead of having the old keyblocks continue with their old seeds, we may want to reseed them like this:

```C
randomizeBuffer(t->keyblock0[i], seed1, OT_SEEDLEN);
randomizeBuffer(t->keyblock0[i], seed2, OT_SEEDLEN);
r->keyblock0[i] = newBCipherRandomGenByKey(seed1);
resetBCipherRandomGen(t->keyblock0[i], seed2);
```

The same for `keyblock1`, and the sender side. It requires a full security proof, which I have not yet done, but I'm fairly confident this tweak will fix it. I'll try to fill out the proof when I can, but feel free to have a swing if you have time. It may be a good excuse for you to learn proving MPC protocols if you haven't picked it up yet ;)

This may or may not make the whole nonce thing redundant, I haven't decided.

src/ext/oblivc/ot.c
```diff
@@ -1033,6 +1080,69 @@ bcipherCryptNoResize(BCipherRandomGen* gen,const char* key,int nonce,
   randomizeBuffer(gen,dest,n);
   memxor(dest,src,n);
 }
+
+
+/*
```
samee: Ok, this is starting to get on my OCD :) It just so happens that recent versions of git highlights trailing whitespaces, and you tend to leave one after each "/*" comment starts.

src/ext/oblivc/ot.c
```diff
@@ -1033,6 +1080,69 @@ bcipherCryptNoResize(BCipherRandomGen* gen,const char* key,int nonce,
...
+  {
+    t->nonce_master = malloc(sizeof(*t->nonce_master));
+    t->ref_count = malloc(sizeof(*t->ref_count));
+    *t->ref_count = 1;
+  }
+  else
+  {
+    t->nonce_master = t_prev->nonce_master;
+    t->ref_count = t_prev->ref_count;
+    __atomic_add_fetch(t->ref_count, 1, __ATOMIC_RELAXED);
```
samee: I don't remember if I've already said this. Can we try C11 atomics instead of GNU extensions? Or do they not work for some reason?

src/ext/oblivc/ot.c
```diff
@@ -1033,6 +1080,69 @@ bcipherCryptNoResize(BCipherRandomGen* gen,const char* key,int nonce,
...
+
+/* 
+   Data structures and functions for dealing with multithreaded nonce-allocation
+   We use gcc's built-in atomics rather than synchronization primitives. On modern CPUs
+   these translate directly into instructions, so they're pretty quick. The reason for
+   all this complication: setting up OT extensions afresh is really expensive, so we
+   want to be able to reuse a single instance across multiple threads
+   */
+
+typedef struct HonestOTExtNonceTracker
```
samee: (optional) Since it's just two pointers, can't we incorporate this directly in `SendMsgArgs` below? You can leave it as is if you think this version is readable, but the extra indirection is kind of bugging me.

samee: Actually, an alternate suggestion would be to rename it to have it not just be a nonce tracker, but any future variable that requires global state across threads, split out of the same `{Send,Recv}MsgArgs` object. Not sure what that name would be though. Not global, since we can have other independent objects in flight. `{Send,Recv}MsgArgsShared`? I guess we will just have to rely more on comments than variable names.

src/ext/oblivc/ot.c
```diff
@@ -1049,7 +1159,9 @@ bcipherCryptNoResize(BCipherRandomGen* gen,const char* key,int nonce,
 typedef struct
 { BCipherRandomGen *cipher;
   const char *box;
-  int n, rowBytes, *rows, k, nonce, nonceDelta, len, destParty, c;
+  HonestOTExtNonceTracker *nonce_trk;
+  size_t nonce, nonceDelta;
```
samee: :+1: Thanks for the type fixes.

samee (commented):
Only one major concern, and the rest are all minor convention-related issues.

src/ext/oblivc/ot.c
jackdoerner: You're welcome. I've been trying to do 32->64 updates whenever I touch things that need them. Someone should probably take a more systematic look at things at some point in the future.

src/ext/oblivc/ot.c
jackdoerner: We can use C11 atomics if you like, but it's worth noting that they're only supported for GCC >= 4.9, whereas the GCC builtins go back to 4.1. Do you care about support for those versions?

src/ext/oblivc/ot.c
jackdoerner: I kept the indirection so that I could separate the atomics and memory-management logic from the main OT extensions logic. I do see your point though. I think I am more in favor of your second proposal, so barring further comment, that is what I will do, but my opinion is not strong, so if you'd really rather go with the first, I am happy to change it to work that way.

src/ext/oblivc/ot.c
jackdoerner: Hmm. I wonder how this happens. I am not doing it consciously.

src/ext/oblivc/ot.c
jackdoerner: Maybe? Should this be a matter of convention (i.e. communicating in the fn signature that the passed struct won't be altered) or of admissibility (anything that can be const should be). If the latter, I suspect there are a lot of things that aren't currently const but should be. If the former, I would argue that reading from the keyblock probably doesn't count (in terms of what we're trying to communicate) as a change, but resetting it probably does count (which seems to be what we're going to do). Thoughts?

jackdoerner:
I'm realizing only now that I never notified you: I believe your concerns have been taken care of. I will handle all of the other issues we talked about in and around this PR (gate counter improvements, correlated OT) in a separate PR once this one is completed.

jackdoerner:
Btw, I still have two other largish PRs to submit after this one. Any chance we can get it merged before too long?

samee:
Ah, thanks for the bump. Setting a reminder on my calendar to do it by next weekend.

PR #52, opened 2018-02-18
Updated README for Fedora 27

jackdoerner:
Fedora seems not to require opam (though in F27, an opam package is available). Tested on a reasonably fresh installation.


samee:
Awesome, thanks for testing.

PR #53, opened 2018-02-26
Fix #38 by rewriting memxor

### LINKED ISSUE #38: Crashes when feeding arrays of sizes not divisible by word sizes
Consider this minimal example:
```C
#include "obliv.oh"
#include "obliv.h"
#include <stdio.h>

#define myint int
#define feedOblivMyintArray feedOblivIntArray

void circuit(void *_arg) {
    size_t k = 101;
    myint *array = calloc(k, sizeof(myint));
    obliv myint *obliv_array = calloc(k, sizeof(obliv myint));
    feedOblivMyintArray(obliv_array, array, k, 2);
    free(array);
    free(obliv_array);
}

int main(int argc, char** argv) {
    int party = atoi(argv[1]);
    printf("sizeof(int) is %zd\n", sizeof(int));
    ProtocolDesc pd;
    if(party == 1) {
        protocolAcceptTcp2P(&pd, "23456");
    } else {
        protocolConnectTcp2P(&pd,"localhost", "23456");
    }
    setCurrentParty(&pd, party);
    execYaoProtocol(&pd, circuit, NULL);
    return 0;
}
```
For convenience, I also put it in a [repo](https://github.com/schoppmp/oblivc-bug) that you can clone and `make run`.

This program runs fine when I do either one or a combination of those:
- `#define myint long` and `#define feedOblivMyintArray feedOblivLongArray`
- `#define myint bool` and `#define feedOblivMyintArray feedOblivBoolArray`
- set `k` to a value such that `k * sizeof(myint)` is divisible by 8

However, it crashes on my machine (Linux 4.9.39-1-MANJARO, Intel(R) Core(TM) i7-3520M) if I do none of those, or if I change `myint` to an even smaller integer type. But then again, this code runs fine on another machine running Ubuntu.

If I run the binary with valgrind, I get the following output:
```
==16549== Memcheck, a memory error detector
==16549== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==16549== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==16549== Command: ./test 1
==16549== 
==16549== 
==16549== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==16549==  General Protection Fault
==16549==    at 0x11D01B: senderExtensionBox (in /tmp/oblivc-example/test)
==16549==    by 0x123528: honestOTExtSend1Of2Start (in /tmp/oblivc-example/test)
==16549==    by 0x1235DF: honestWrapperSend (in /tmp/oblivc-example/test)
==16549==    by 0x10B7F3: yaoGenrFeedOblivInputs (in /tmp/oblivc-example/test)
==16549==    by 0x113907: feedOblivIntArray (in /tmp/oblivc-example/test)
==16549==    by 0x109F90: circuit (in /tmp/oblivc-example/test)
==16549==    by 0x10E561: execYaoProtocol (in /tmp/oblivc-example/test)
==16549==    by 0x10A0E9: main (in /tmp/oblivc-example/test)
==16549== 
==16549== HEAP SUMMARY:
==16549==     in use at exit: 319,524 bytes in 205 blocks
==16549==   total heap usage: 229,805 allocs, 229,600 frees, 7,556,645 bytes allocated
==16549== 
==16549== LEAK SUMMARY:
==16549==    definitely lost: 32,320 bytes in 1 blocks
==16549==    indirectly lost: 0 bytes in 0 blocks
==16549==      possibly lost: 0 bytes in 0 blocks
==16549==    still reachable: 287,204 bytes in 204 blocks
==16549==         suppressed: 0 bytes in 0 blocks
==16549== Rerun with --leak-check=full to see details of leaked memory
==16549== 
==16549== For counts of detected and suppressed errors, rerun with: -v
==16549== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

I tried finding out where exactly things go wrong, but the bug disappears as soon as I compile Obliv-C with `-g` :/

Any ideas? My guess is that Obliv-C reads inputs in words and this leads to a segfault at the last word (which isn't completely allocated), but I haven't been able to find the exact location where things go wrong.

schoppmp:


src/ext/oblivc/obliv_common.h
```diff
@@ -126,24 +126,32 @@ static inline void setBit(char *dest,int ind,bool v)
...
+  n /= sizeof(uint64_t);
+  uint64_t *ldest = (uint64_t*) dest;
+  uint64_t *lsrc = (uint64_t*) src;
+  // make sure n is a multiple of 128 bits
+  if (n & 1) {
+    n--;
+    ldest[n] ^= lsrc[n];
+  }
+  // process the rest in batches of two uint64_t,
+  // allowing the compiler to use 128-bit pxor (SSE2)
```
samee: Have you checked with -S if the compiler actually does this?

src/ext/oblivc/obliv_common.h
```diff
@@ -126,24 +126,32 @@ static inline void setBit(char *dest,int ind,bool v)
...
+{
+  // xor the last n % 8 bytes; Compiler optimization will unroll this loop
+  size_t offset = n % sizeof(uint64_t);
+  for(size_t i = n - offset; i < n; i++) {
+    ((char *)dest)[i] ^= ((char *)src)[i];
   }
-  if(i+sizeof(uint16_t)<=n)  {
-    *((uint16_t*)((char*)dest+i)) ^= *((uint16_t*)((char*)src+i));
-   i+=sizeof(uint16_t);
+  n /= sizeof(uint64_t);
```
samee: I realize that I am asking for microoptimizations here, but does it make any performance difference to start this wide xor from a 16-byte boundary?

samee (approved):
Two minor comments, but this mostly looks good.

schoppmp:
I think this issue is a good example of premature optimization on our part. The following already gives a [highly optimized assembly](https://github.com/samee/obliv-c/files/1767007/memxor.txt):
```C
void memxor(void *restrict dest, const void *restrict src, size_t n) {  
  for(size_t i = 0; i < n; i++) {
    ((char *) dest)[i] ^= ((char *)src)[i]; 
  }
}
```
Most of it is reduced to a `pxor` loop (`.L6`, lines 95–102).
Note that the assumption that `src` and `dest` do not overlap, which allows us to use `restrict` pointers, was already implicitly made by all our hand-written versions.
In my opinion, the above is what we should use.

schoppmp:
@samee Did you get a chance to try and see if the simplified loop with restrict pointers compiles to the same optimized assembly for you?

samee:
Oh I didn't realize you were waiting for me. The actual assembly seems to vary a bit depending on inlining site, but I agree with you that it's already doing a fair bit of vectorizing in most cases. Plus, the code is simpler. Unless somebody comes up with a strong performance regression numbers, I am definitely okay with this change.

samee:
Although, there *are* some weird cases where it's not vectorizing. (e.g. in psi.c:sendHashes, gcc 7.2.0). Not sure why.

PR #57, opened 2018-03-23
Support large batch OT size for `honestOTExtSend1Of2Chunk`

weikengchen:
This PR changes some `int`-type size to `size_t`-type size in the OT for `honestOTExtSend1Of2Chunk`.

Therefore, it can support large batch OT size. 

Previously, the size may go to a negative int before calling `malloc`. 
When it converts to `size_t`, it is negative, too. 
So if you ask for 4GB, it becomes several EBs. 

This PR also resolves the failure of Square-Root ORAM for large block sizes.
More details of that issue: https://github.com/samee/sqrtOram/issues/3

Tested with floram sqrtOram benchmark.

### DIFF
--- a/bin/oblivcc
+++ b/bin/oblivcc
@@ -3,6 +3,7 @@ OCCPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../ && pwd )"
 OCLIB=$OCCPATH/src/ext/oblivc
 COMPILE_ONLY=0
 CCARGS=""
+AFTERARGS=""
 CILLYARGS=""
 OCSRC=()
 GCRYPT="-lgcrypt"
@@ -12,26 +13,15 @@ else
   OBLIV_BITS="-include obliv_bits.h"
 fi
 
-while [ $# -ge 1 ]; do
-  if [ "$1" = "-static-gcrypt" ]; then
-    GCRYPT="/usr/lib/x86_64-linux-gnu/libgcrypt.a -lgpg-error"
-  else
-    CCARGS+="$1 "
-  fi
-  shift
-done
-
 set -e
-$OCCPATH/bin/cilly -Wno-format-security -Wno-deprecated-declarations --keepunused --doprocessObliv --markImplicitCasts $OBLIV_BITS -I $OCLIB -L$OCCPATH/_build -lobliv $GCRYPT -pthread $CCARGS
 
-if false; then
 while [ $# -ge 1 ]; do
   if [ "$1" = "-c" ]; then
     COMPILE_ONLY=1
     CCARGS+="-c "
   elif [ "$1" = "-static-gcrypt" ]; then
     GCRYPT="/usr/lib/x86_64-linux-gnu/libgcrypt.a -lgpg-error"
-  elif [ "${1%.oc}" != "$1" ]; then
+  elif [[ "$1" == *oc ]]; then
     OCSRC+=($1)
     CCARGS+="$1.cil.c "
   else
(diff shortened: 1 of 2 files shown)


samee (commented):
Would it be slightly better to change MSGBUFFER_SIZE to be defined as ((size_t)10000), instead of having to change the constant usage everywhere?

weikengchen:
I have another idea, change the `len` in the data structure?

change `bufused`, `len` (this one) in SendMsgArgs,
and `bufread`, `payloadLeft`, `len` (this one) in RecvMsgArgs?

This can reduce the range of changes to only the definitions of the data structure.
Together with the changes for `bufused`, `bufread`, `payloadLeft`.

weikengchen:
This practice would be safer, somehow.

As if somewhere else uses it, it must use `len`.

weikengchen:
Like, the code Ln 1131, changing MSGBUFFER_SIZE may not be helpful, but `a->len` can.

`transRecv(a->trans,a->srcParty,a->buf,n*(size_t)(a->len));`

I will push another commit.

weikengchen:
OK. I have made the change.

Now I will test the change on sqrtOram.

weikengchen:
Pass. 

@samee  please comment :)
And thanks a lot for the obliv-c, sqrtOram, and half-gate!

samee:
I certainly agree this is much cleaner. But if I may, I'd like to enforce a bit of reasoning behind the choices of types. E.g. when writing or changing code in `ot.c`, I no longer have any sense or intuition of which variables (both in the struct and in function local scope) should be `size_t` vs `int`. We could certainly throw our hands up and turn them all into size_t, but remember that size_t doesn't guarantee 64-bits. Instead, let's see if we can reason around this.

At the OT abstraction, what are we trying to support here? Billions of messages, each of them only a few bytes, or a small number of messages, each of them a billion bytes? If we want a large number of messages, notice that types like `nonce` may start silently repeating, breaking security.

Possible variants I am thinking of: maybe we should just use uint64_t instead of size_t, perhaps typedefed to our own size type. Maybe we should have a medium stress-test that checks if nonces ever repeat. But a lot of these will depend on the particular application we are trying to support, so maybe you can tell me a bit more about how you are exercising sqrtOram, and how that affects these various struct members.

Sorry if this is making a potentially simple fix into something more involved, but I'd like to avoid silently breaking security assumptions :)

weikengchen:
I will prefer to send a description of the application via email.

Do you still use `sza4uq@virginia.edu`?

samee:
Sounds good. Yes, that email address still works.

weikengchen:
side questions:

I may set the MSGBUFFER_SIZE smaller to see if sqrtOram can live even longer. 

What will happen if MSGBUFFER_SIZE is too small? So the buffer is not that sufficient.  I kind of feel that it might still be ok.

Or maybe, I can slow down the OT to wait for the buffer. 

Why the number 10000 was selected here at that time?

samee:
This entire `ot.c` was written to support Yao's garbled circuits. So OT was mostly targeted towards transferring 80-bit labels. So len=80/8=10 bytes in most intended cases. Since you mentioned SqrtORAM, I think that is also the case there ... although I could be wrong. You can try debug-printing `len` to see what value is being used in your case.

10000 was selected empirically, and you can certainly try adjusting that. If you make it too small, the buffer will fill up quickly and we will perform expensive I/O operations too frequently. If it is too large, we will have a large memory buffer and you may run out of memory. Plus, if it is too large, it can also hurt performance in the form of increased latency.

weikengchen:
Wrong commits are added here. Sorry

I would close this first.

PR #61, opened 2018-04-13
Fix minor build script bug

mark-i-m:
It turns out that if `CDPATH` is set, `cd` always outputs the cwd, which messes up some paths in the oblivcc script. This simply pipes the output of `cd` to `/dev/null` so that we only get the output of `pwd` in the variable.


samee:
Thanks for the patch!

Issue #59, opened 2018-04-05
`processObliv.ml` cannot handle C labels, like `__gmp_done` in <gmp.h>


weikengchen:
This is relevant to https://github.com/samee/obliv-c/issues/58.

## Background
The issue shows that labels are discarded magically, not by CIL, but by Obliv-C's modification.

We know `oblivcc` processes all C/OC files with `cilly`. The main change from `CIL` to `Obliv-C` backend is `processObliv.ml`. After the test,  we find that the label is discarded by`processObliv.ml`.

We know for sure that it is not practical to consider `goto` in an `obliv if`, but in non-obliv areas, people may import headers from other libraries. 

Although `goto` is strongly not encouraged nowadays for daily programming, some third-party libraries, like GMP (<gmp.h>) uses `goto` -- this will cause compile errors in Obliv-C.

## Reason
My first time OCaml :)

After a careful debug, we find the following:

`codegenVisitor` rewrites all statements. When it matches `Instr ilist` where `ilist` should be a list, it calls `codegenInstr` to rewrite. This one does not produce a successful new statement for C labels, in which `List.length ilist` is 0 and you cannot reproduce a statement from `ilist`. 

```
    | Instr ilist -> 
        let nestedGen = codegenInstr curCond tmpVar isDeepVar in
        ChangeTo (mkStmt (Instr (mapcat nestedGen ilist)))
```

Very special the way CIL treats labels. The `vstmt` on the labels also has no line number and filename associated with it when we use `E.warn` to debug. 

## Solution
It can be fixed by `DoChildren` instead of `ChangeTo` when `ilist` is empty.
```
    | Instr ilist -> 
        let num = List.length ilist in
        let nestedGen = codegenInstr curCond tmpVar isDeepVar in
        if num == 0 then DoChildren
        else ChangeTo (mkStmt (Instr (mapcat nestedGen ilist)))
```

(A fixed version of processObliv.ml: https://gist.github.com/weikengchen/ed7ec649fbac17b67df5cce7abf60194)

## Side note on `goto` in `obliv` areas
Should we prohibit `goto` in `obliv`?

It will be adding in `class controlCheckVisitor` with `gotoOk` similar to `returnOk`, with the following `E.error`:

```
    | Goto (_,loc) -> if not !gotoOk then
             E.s (E.error "%a: unexpected goto" d_loc loc)
```

## Question
need a PR?

samee:
[...]

PR #62, opened 2018-04-16
fix `processObliv` dropping labels from CIL && ||

weikengchen:
CIL may generate `goto` and labels from && || in some cases (see https://people.eecs.berkeley.edu/~necula/cil/cil004.html).

However, due to an implementation detail, `processObliv` drops the label, causing `gcc`'s compiler error.

This fix corrects the `codegenVisitor` when dealing `statement` with empty list. Such statement cannot be regenerated by the information in the list, and should not be reconstructed.

### DIFF
--- a/src/ext/processObliv/processObliv.ml
+++ b/src/ext/processObliv/processObliv.ml
@@ -802,8 +802,10 @@ class codegenVisitor (curFunc : fundec) (dt:depthTracker) (curCond : lval)
     let isDeepVar v = dt#curDepth() = dt#varDepth v in
     match s.skind with
     | Instr ilist -> 
+        let num = List.length ilist in
         let nestedGen = codegenInstr curCond tmpVar isDeepVar in
-        ChangeTo (mkStmt (Instr (mapcat nestedGen ilist)))
+        if num == 0 then DoChildren
+        else ChangeTo (mkStmt (Instr (mapcat nestedGen ilist)))
     | If(c,tb,fb,loc) when isOblivBlock tb ->
         let cond = "__obliv_c__cond" in
         let cv = var (tmpVar ~name:cond oblivBoolType) in
--- a/src/ext/processObliv/processObliv.ml
+++ b/src/ext/processObliv/processObliv.ml
@@ -802,10 +802,12 @@ class codegenVisitor (curFunc : fundec) (dt:depthTracker) (curCond : lval)
     let isDeepVar v = dt#curDepth() = dt#varDepth v in
     match s.skind with
     | Instr ilist -> 
+        let mkStmt_with_label (sk: stmtkind) (lb: label list) : stmt =
+        { skind = sk; labels = lb; sid = -1; succs = []; preds = [] } in
         let num = List.length ilist in
         let nestedGen = codegenInstr curCond tmpVar isDeepVar in
         if num == 0 then DoChildren
-        else ChangeTo (mkStmt (Instr (mapcat nestedGen ilist)))
+        else ChangeTo (mkStmt_with_label (Instr (mapcat nestedGen ilist)) s.labels)
     | If(c,tb,fb,loc) when isOblivBlock tb ->
         let cond = "__obliv_c__cond" in
         let cv = var (tmpVar ~name:cond oblivBoolType) in


weikengchen:
This is relevant to https://github.com/samee/obliv-c/issues/59.

https://github.com/samee/obliv-c/issues/59 also proposed a naïve check for `goto` inside `obliv c`. As Samee mentioned, that naïve check does not work -- they do not play together with CIL generating `goto` and `label` in those cases. The naïve check is not included in this PR.

So, this PR only fixes `labels drop`.

samee:
Slow down:) When it comes to programming language or security, it makes sense to think through all possible corner cases and see if it helps.
 
Can you tell me what effect this patch will have on code like this:

    if (f1() && f2()) {...}

Here, let's say f1() returns obliv bool, and f2() modifies non-obliv variables. Should that be allowed? Will the user expect that to work, or be surprised by it? How will it work if || replaces &&? I completely agree that these cases need to be solved, but let's break these problems down one at a time. Let me comment on the goto issue #59 first, and then we will come back to this one.

weikengchen:
Let us work on #59 first. 

This PR is not for new features.

weikengchen:
OK. So this one is relevant to the issue #63 

You mentioned the case:
```
if (f1() && f2()) {...}
```

But, does it make sense that `f1()` returning an oblivious bool can be a condition in non-obliv environment?
We need to reveal it first. 

If what you mean is:
```
obliv if (f1() && f2()) {...}
```

I feel that we are required to use `&`?
```
obliv if (f1() & f2()) {...}
```

weikengchen:
FYI.

I tried some cases you mentioned to see if Obliv-C has already limited them.

If we use obliv bool in a non-obliv condition:
```
int func(){
        return 1;
}

int main(){
  obliv bool a = true;
  obliv bool b = false;
  obliv bool c = true;

  if(a && func()){
       b = false;
  }

  return 0;
}
```

The compiler says: 
```
Error: ./test.c:11:Cannot use obliv-type expression as a condition
Fatal error: exception Errormsg.Error
```

If I use `&&` or `||`,
```
int main(){
  obliv bool a = true;
  obliv bool b = false;
  obliv bool c = true;

  obliv if(a && b){
       c = false;
  }

  return 0;
}
```

The compiler says: 
```
Error: ./test.c:6:Cannot use obliv-type expression as a condition
Fatal error: exception Errormsg.Error
```

weikengchen:
And the case you mentioned:

```
obliv bool func(){
        // do some modifications
        return true;
}

int main(){
  obliv bool a = true;
  obliv bool b = false;
  obliv bool c = true;

  obliv if(a & func()){
       c = false;
  }

  return 0;
}
```

It seems that `func()` will run.

```
  {
# 6 "./test.c"
  memset((void *)(& a), 0, sizeof(__obliv_c__bool ));
# 6 "./test.c"
  memset((void *)(& b), 0, sizeof(__obliv_c__bool ));
# 6 "./test.c"
  memset((void *)(& c), 0, sizeof(__obliv_c__bool ));
# 6 "./test.c"
  memset((void *)(& tmp), 0, sizeof(__obliv_c__bool ));
# 7 "./test.c"
  __obliv_c__setSignedKnown(& a, 1UL, (widest_t )((_Bool )1));
# 8 "./test.c"
  __obliv_c__setSignedKnown(& b, 1UL, (widest_t )((_Bool )0));
# 9 "./test.c"
  __obliv_c__setSignedKnown(& c, 1UL, (widest_t )((_Bool )1));
# 11 "./test.c"
  tmp = func();
  {
# 11 "./test.c"
  __obliv_c__setZeroExtend(& __cil_tmp5, 32UL, & tmp, 1UL);
# 11 "./test.c"
  __obliv_c__setZeroExtend(& __cil_tmp6, 32UL, & a, 1UL);
# 11 "./test.c"
  __obliv_c__setBitwiseAnd(& __cil_tmp7, & __cil_tmp6, & __cil_tmp5, 32UL);
  {
```

weikengchen:
A new commit follows. This commit totally fixes this problem.

Problem: When processObliv reproduces a statement, the label is not copied.

Will close the issue #62.

samee:
Ok thanks a lot. This looks great, and I'll merge it in. Can you quickly update this thread with what case is covered by the `num == 0` check? I still can't remember what kind of code causes that to become true, and I'm sure I'll forget it again.

weikengchen:
Morning!

Previously, a partial solution (https://github.com/samee/obliv-c/issues/59) chooses not to rewrite a statement when `num == 0`, in order to solve the case in `__gmp_done`.

However, this is not the general solution. This solution works because:

```
#define __GMPN_AORS(cout, wp, xp, xsize, yp, ysize, FUNCTION, TEST)     \
  do {                                                                  \
    mp_size_t  __gmp_i;                                                 \
    mp_limb_t  __gmp_x;                                                 \
                                                                        \
    /* ASSERT ((ysize) >= 0); */                                        \
    /* ASSERT ((xsize) >= (ysize)); */                                  \
    /* ASSERT (MPN_SAME_OR_SEPARATE2_P (wp, xsize, xp, xsize)); */      \
    /* ASSERT (MPN_SAME_OR_SEPARATE2_P (wp, xsize, yp, ysize)); */      \
                                                                        \
    __gmp_i = (ysize);                                                  \
    if (__gmp_i != 0)                                                   \
      {                                                                 \
        if (FUNCTION (wp, xp, yp, __gmp_i))                             \
          {                                                             \
            do                                                          \
              {                                                         \
                if (__gmp_i >= (xsize))                                 \
                  {                                                     \
                    (cout) = 1;                                         \
                    goto __gmp_done;                                    \
                  }                                                     \
                __gmp_x = (xp)[__gmp_i];                                \
              }                                                         \
            while (TEST);                                               \
          }                                                             \
      }                                                                 \
    if ((wp) != (xp))                                                   \
      __GMPN_COPY_REST (wp, xp, xsize, __gmp_i);                        \
    (cout) = 0;                                                         \
  __gmp_done:                                                           \
    ;                                                                   \
  } while (0)
```

`__gmp_done` is at the end of a `{}`, therefore, it will be a single statement, with a label, but contains no code (`num == 0`).

By not rewriting the code when `num == 0`, this special case is solved.

Yet, as I mentioned, it is not the general case. Consider another header file <fcntl.h>, which can be used to turn a socket into `non-blocking` and allows `online pre-computation` when one party is waiting for another party's response.

<fcntl.h> may include <fcntl2.h>, which has some `if` conditions that will be optimized by CIL into shortcut `goto`:

```
if (__builtin_constant_p (__oflag))
    {
      if ((__oflag & O_CREAT) != 0 && __va_arg_pack_len () < 1)
	{
	  __openat64_missing_mode ();
	  return __openat64_2 (__fd, __path, __oflag);
	}
      return __openat64_alias (__fd, __path, __oflag, __va_arg_pack ());
}
```
where CIL will create `_L` label for `if ((__oflag & O_CREAT) != 0 && __va_arg_pack_len () < 1)`.

In this case, in the rewrite code, `_L` label is in a statement with some codes,

```
      _L:
      __openat64_missing_mode ();
      ......
```
which is no longer a statement with `num == 0`.


Therefore, `num == 0` is not the general fix to the problem. 

The problem should be: 

> for any statement that has labels, `processObliv` forgets to copy the labels when rewriting.

The previous solution (wrong), which is:

> assumed labels are zero-code statements (wrong), we can choose not to process them in `processObliv`.

Now, the solution is:

> for any statement that has labels, `processObliv` NOW copies the labels when rewriting.

samee:
Thanks a lot for the detailed explanation. I tried to simplify the logic a bit before merging in. Let me know if it broke anything on your side.

weikengchen:
will have a try on the `gmp.h` case.

Side note: if you do not check `num == 0`, then you don't need to `let num = ...`

weikengchen:
oh. just found that you have already deleted `let num = ...`

weikengchen:
pass `gmp.h` and `fcntl.h` test.

Great job!

PR #65, opened 2018-06-19
Fix C++ compatibility issue

schoppmp:
Remove the `restrict` keyword if obliv_common.h is included in C++ code. Ideally, that would never be needed in the first place, but there are some strange dependencies (e.g. via bcrandom.h)

### DIFF
--- a/src/ext/oblivc/obliv_common.h
+++ b/src/ext/oblivc/obliv_common.h
@@ -130,7 +130,13 @@ static inline bool getBit(const char* src,int ind)
   { return src[ind/8]&(1<<ind%8); }
 static inline void xorBit(char *dest,int ind,bool v)
   { dest[ind/8]^=(v<<ind%8); }
+
+// C++ doesn't have the restrict keyword
+#ifdef __cplusplus
+static inline void memxor(void *dest, const void *src, size_t n) {
+#else
 static inline void memxor(void *restrict dest, const void *restrict src, size_t n) {
+#endif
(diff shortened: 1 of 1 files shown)


samee:
Can you leave a comment on why it was required to begin with? I just ran a quick `make`, nothing seemed to break. Looks like you introduced it back in https://github.com/samee/obliv-c/commit/5f9620706424aa121a490719b9be3099042f4625.

schoppmp:
Yes, the restrict pointers were introduced as part of #53. They allow the compiler to optimize more, since they assert that the memory regions passed to `memxor` do not overlap. I just tested it again with GCC 8.1.1 and the assembly is indeed shorter. Not sure how much performance this saves in practice, though. For that we would need proper benchmarks...

samee:
Interesting. We could always use the preprocessor for this:

```
#ifdef __cplusplus
#define RESTRICT
#else
#define RESTRICT restrict
#endif
```

Btw, how does this file get included into c++ anyway? Should we refactor dependencies?

weikengchen:
glad to see Obliv-C++

schoppmp:
> Btw, how does this file get included into c++ anyway? Should we refactor dependencies?

It gets included in [bcrandom.h](https://github.com/samee/obliv-c/blob/obliv-c/src/ext/oblivc/bcrandom.h#L5), which I then included from C++ myself. There's already a comment there about refactoring dependencies.

PR #79, opened 2018-12-17
fix macos build problem related to finding libgcrypt

justin1121:
Hello,

This fixes a problem with building obliv-c for macos where it couldn't find the libgcrypt library. I ended up having to modify the Makefile.in file but I'm pretty sure that is suppose to be generated from another file such as a Makefile.am but couldn't find one. If this fix doesn't make sense let me know how I can fix this properly.

Thanks!

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -37,6 +37,7 @@ stamp-h: config.h.in config.status
 
 CC=@CC@
 CPP=@CPP@
+CFLAGS=@CFLAGS@
 @DEFAULT_COMPILER@=1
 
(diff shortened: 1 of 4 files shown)


samee:
Thanks for your changes. I agree it is weird to have a file with `@do_not_edit@` tags to be directly edited. Looking at `git log Makefile.in` I see that the file has a long history of being directly edited, so I don't think you are violating any existing conventions. This file was inherited from https://github.com/cil-project/cil, so maybe somebody there knows the history behind it.

PR #88, opened 2019-02-20
Fix assignment of clear float values to obliv floats

### LINKED ISSUE #85: obliv float not working with constants
Hi,
I tried the following code and it seems that obliv float data type has some issue with constant float assignment
```obliv float s = 1.1;```
```revealOblivFloat(&io->result,s,0);```

I am getting result = 1.0000

and if the contant float between 0 and 1 the result is always 0. any suggestions?

### PR DESCRIPTION
**Pull request by schoppmp**
Casting to `widest_t` seems truncate the inputs. An example that did not work before this change:
```
obliv float z = 0.5f;
float z_clear;
revealOblivFloat(&z_clear, z, 0);
printf("expected 0.5, got %f\n", z_clear);
```

Output before:
```expected 0.5, got 0.000000```

Output now:
```expected 0.5, got 0.500000```

Fixes #85

### DIFF
--- a/src/ext/oblivc/obliv_bits.h
+++ b/src/ext/oblivc/obliv_bits.h
@@ -192,9 +192,9 @@ void __obliv_c__condAssignKnown(const void* cond, void* dest, size_t size
 // Conditionals (TODO other operators) that may be faster at times
 static inline 
 void __obliv_c__condAssignKnownF(const void* cond, void* dest, size_t size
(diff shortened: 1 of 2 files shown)


samee:
Thanks. Casting truncates output simply because it is floating point to integer type conversion. That would never have worked unless it went through some weird bit-manipulation magic.

samee:
Merged.

PR #98, opened 2021-03-01
updates macos installation instructions

wrharris:
Adds instructions for installing on macOS 11 (Big Sur).


samee:
This is awesome, thanks!

PR #99, opened 2021-06-18
added caveat to macOS build instructions

mitchnegus:
Adds a caveat to the README build instructions for macOS for instances where only the Xcode command line tools are installed (not the full version of Xcode). Since I also just ran this process (mostly successfully) on macOS 10.15, I included 10.15 (Catalina) alongside 11 in the installation instructions.

Note: Source for the Xcode command line tools adjustment was found in [nalzok's comment on this StackOverflow post](https://stackoverflow.com/a/58323411)


samee:
Thanks for trying it out!

